PR #7, opened 2021-05-25
use multistage dockerfile to pull consul-ecs from releases

alvin-huang:
This trues up the Dockerfile in this repo with how we normally build docker containers. It uses a mult-stage docker build to pull the `consul-ecs` binary down and verifies GPG signatures, as well as applies a `maintainer` and `version` label which we use to check what the latest release is behind the `latest` docker tag.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,7 +7,7 @@ ENV HASHICORP_RELEASES=https://releases.hashicorp.com
 
 # Set up certificates, base tools, and software.
 RUN set -eux && \
-    apk add --no-cache ca-certificates curl gnupg libcap openssl su-exec iputils libc6-compat iptables && \
+    apk add --no-cache ca-certificates curl gnupg && \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,3 +1,55 @@
+FROM alpine:latest as BUILDER
+ARG NAME=consul-ecs
+ARG VERSION
+
+# This is the location of the releases.
+ENV HASHICORP_RELEASES=https://releases.hashicorp.com
+
+# Set up certificates, base tools, and software.
+RUN set -eux && \
+    apk add --no-cache ca-certificates curl gnupg libcap openssl su-exec iputils libc6-compat iptables && \
```
lkysow, line 10: do we need all of these in the builder image? Probably don't need su-exec, iputils, iptables

alvin-huang: oh yeah I just copied that over, let me minimize this...

PR #11, opened 2021-07-12
mesh-init: Configure Envoy to talk over TLS to the Consul client

ishustava:
This PR proposes the following changes:
- Change mesh-init command to configure Envoy to use TLS with a trusted CA cert. We're using a trusted CA here because currently Consul doesn't allow you to generate Envoy bootstrap json with TLS verification skipped, and we didn't want to modify the Envoy json ourselves.
- Add `dev-docker` to `Makefile`
- fix dockerignore to exclude the right binary so that building a docker image locally works.


pglass (approved):
Looks good to me! Tests passed locally.

Couple of nit picks - not important.

subcommand/mesh-init/command_test.go
```diff
@@ -0,0 +1,171 @@
...
+			}))
+			t.Cleanup(ecsMetadataServer.Close)
+			os.Setenv(awsutil.ECSMetadataURIEnvVar, ecsMetadataServer.URL)
+
+			ui := cli.NewMockUi()
+			cmd := Command{
+				UI: ui,
+			}
+
+			envoyBootstrapFile, err := ioutil.TempFile("", "")
```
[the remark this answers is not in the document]
ishustava: Great catch! will update it!

subcommand/mesh-init/command_test.go
```diff
@@ -0,0 +1,171 @@
...
+			expectedServiceRegistration := &api.AgentService{
+				ID:      "test-service-abcdef",
+				Service: "test-service",
+				Port:    c.servicePort,
+				Meta: map[string]string{
+					"task-id":  "abcdef",
+					"task-arn": "arn:aws:ecs:us-east-1:123456789:task/test/abcdef",
+					"source":   "consul-ecs",
+				},
+			}
```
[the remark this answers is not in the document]
ishustava: yep, good idea!

subcommand/mesh-init/command.go
```diff
@@ -154,14 +158,57 @@ func (c *Command) realRun(log hclog.Logger) error {
 
 	log.Info("service and proxy registered successfully", "name", serviceName, "id", serviceID)
 
+	// Grab CA root from the Consul server if TLS is enabled.
```
lkysow, line 161: Why not disable CA verification for Envoy since it's talking localhost to Consul so there's no MITM risk.

ishustava: How do you disable TLS verification for envoy through the Consul cli? I haven't found anything in the consul codebase that would read `CONSUL_HTTP_SSL_VERIFY` env var for the `consul connect envoy -bootstrap` command or a flag that would do similar for that command.

lkysow: So for the initial call that that command makes itself to Consul, I think it respects the env var. The problem is that the generated bootstrap config doesn't have verification turned off. I edited the generated config in my branch to work around this: https://github.com/hashicorp/consul-ecs/blob/secure/subcommand/mesh-init/command.go#L209-L226

lkysow (approved):
🚀

ishustava:
Closing this PR as we've decided to not have TLS for HTTP and gRPC for the consul client.

I've moved the unit tests and other fixes to #18.

PR #19, opened 2021-07-26
Unit tests for discover-servers

pglass:


subcommand/discover-servers/command_test.go
```diff
@@ -0,0 +1,104 @@
+package meshinit
```
ishustava, line 1:
```suggestion
package discoverservers
```

subcommand/discover-servers/command.go
```diff
@@ -10,6 +10,7 @@ import (
 	"github.com/aws/aws-sdk-go/aws"
```
ishustava, line 10: we also need to change package name on line 1.

subcommand/discover-servers/command_test.go
```diff
@@ -0,0 +1,104 @@
...
+			return nil
+		},
+		"no-containers": func(task *ecs.Task) []*ecs.Task {
+			task.Containers = nil
+			return []*ecs.Task{task}
+		},
+		"no-network-interfaces": func(task *ecs.Task) []*ecs.Task {
+			task.Containers[0].NetworkInterfaces = nil
+			return []*ecs.Task{task}
+		},
```
ishustava, line 56: should we also test for a case when there is no `PrivateIpv4Address`?

ishustava (commented):
Looks awesome!! I loved how you simulated incomplete tasks.

I left a couple of comments, but once resolved, this is good to go!

subcommand/discover-servers/command_test.go
pglass: Yep, added! Also, added a case where containers are returned, but none are named `consul-server`.

lkysow (approved):
Awesome work!

subcommand/discover-servers/command_test.go
```diff
@@ -0,0 +1,104 @@
...
+			taskMetadataResponse := fmt.Sprintf(`{"Cluster": "test", "TaskARN": "%s", "Family": "test-service"}`, taskARN)
+			ecsMetadataServer := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+				if r != nil && r.URL.Path == "/task" && r.Method == "GET" {
+					_, err := w.Write([]byte(taskMetadataResponse))
+					require.NoError(t, err)
+				}
+			}))
+			t.Cleanup(ecsMetadataServer.Close)
+			os.Setenv(awsutil.ECSMetadataURIEnvVar, ecsMetadataServer.URL)
+
```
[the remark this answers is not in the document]
pglass: Fixed! I actually switched to [`t.Cleanup`](https://pkg.go.dev/testing#T.Cleanup) instead of `defer`, to be consistent.

ishustava (approved):
🎉

PR #20, opened 2021-08-16
Set region for AWS SDK in discover-servers

pglass:


awsutil/awsutil.go
```diff
@@ -29,6 +29,22 @@ func (e ECSTaskMeta) TaskID() string {
...
 }
 
+func (e ECSTaskMeta) Region() string {
+	// Use the region from Task metadata if missing from environment (for EC2).
+	// https://github.com/aws/containers-roadmap/issues/337
+	for _, envKey := range []string{"AWS_REGION", "AWS_DEFAULT_REGION"} {
+		if region := os.Getenv(envKey); region != "" {
+			return region
+		}
+	}
```
lkysow, line 39: It feels a bit weird to me that this method on `ECSTaskMeta` is reaching out to env vars. I could see that be confusing in the future "Why is it returning this region when the metadata json response has _that_ region".

I think a function like:

```go
// Region returns the current AWS region by reading
// region environment variables and falling back to the
// region set in the task metadata.
func Region(meta ECSTaskMeta) string {}
```

Makes more sense to me but it's just an opinion. :) Should wait and see with Iryna thinks too.

awsutil/awsutil_test.go
```diff
@@ -0,0 +1,47 @@
...
+		{
+			name:         "aws-region",
+			env:          map[string]string{"AWS_REGION": "region-from-aws-region"},
+			expectRegion: "region-from-aws-region",
+		},
+		{
+			name:         "aws-default-region",
+			env:          map[string]string{"AWS_DEFAULT_REGION": "region-from-aws-default-region"},
+			expectRegion: "region-from-aws-default-region",
+		},
```
lkysow, line 30: Can you add a test case that if both env vars set it uses AWS_REGION.

awsutil/awsutil_test.go
```diff
@@ -0,0 +1,47 @@
...
+	}
+
+	for _, c := range cases {
+		os.Clearenv()
+		for k, v := range c.env {
+			require.NoError(t, os.Setenv(k, v))
+		}
+
+		ecsMeta := ECSTaskMeta{
```
lkysow, line 39: nit: if this struct was declared at the beginning of the test function then it would make the first test case a bit clearer because I could see where `bogus-east-1` is coming from.

subcommand/mesh-init/command.go
```diff
@@ -72,7 +72,7 @@ func (c *Command) realRun(log hclog.Logger) error {
 	}
 
 	// Register the service.
-	taskID := taskARNToID(taskMeta.TaskARN)
+	taskID := taskMeta.TaskID()
```
lkysow, line 75: 🍺

lkysow (approved):

awsutil/awsutil.go
pglass: Makes sense to me! 

As I'm looking at this again, I'm wondering if I can rely on the SDK to do the environment variable discovery for us.

Then we could have a function that returns a config object for any AWS clients:

```
// ClientConfig returns a prepared config object for AWS clients.
// It pulls the region from the task metadata if not found in the environment.
func ClientConfig(meta ECSTaskMeta) *aws.Config {}
```

pglass: Following my comment above, I added an `awsutil.NewSession` function, which relies on the AWS SDK to find the region from standard places, and falls back on Task meta if no region was found (also, allows us other session setup, like the user-agent handler).

awsutil/awsutil.go
```diff
@@ -60,3 +71,22 @@ func UserAgentHandler(caller string) request.NamedHandler {
...
 }
+
+// NewSession prepares a client session.
+// The returned session includes a User-Agent handler to enable AWS to track usage.
+// If the AWS SDK fails to find the region, the region is parsed from Task metadata
+// (on EC2 the region is not typically defined in the environment).
+func NewSession(meta ECSTaskMeta, userAgentCaller string) (*session.Session, error) {
+	clientSession, err := session.NewSession()
+	if err != nil {
+		return clientSession, err
```
lkysow, line 82:
```suggestion
		return nil, err
```

lkysow: nit: I know it doesn't really matter since go convention says don't use anything if err != nil but makes it a bit clearer it's not to be relied on.

awsutil/awsutil_test.go
```diff
@@ -0,0 +1,76 @@
...
+func TestECSTaskMeta(t *testing.T) {
+	taskRegion := "bogus-east-1"
+	nonTaskRegion := "some-other-region"
+	ecsMeta := ECSTaskMeta{
+		Cluster: "test",
+		TaskARN: fmt.Sprintf("arn:aws:ecs:%s:123456789:task/test/abcdef", taskRegion),
+		Family:  "task",
+	}
+
+	cases := []struct {
```
lkysow, line 21: Does't matter but worth noting another patter you'll see is `map[string]struct` and putting the test name as the key.

lkysow (approved):
Nice! I like the refactor a lot. 🎉

awsutil/awsutil.go
```diff
@@ -29,6 +31,15 @@ func (e ECSTaskMeta) TaskID() string {
 	return split[len(split)-1]
 }
 
+func (e ECSTaskMeta) Region() string {
+	// https://github.com/aws/containers-roadmap/issues/337
+	split := strings.Split(e.TaskARN, ":")
+	if len(split) < 4 {
+		return ""
```
ishustava, line 38: should we return error? If the AWS SDK client will not function without a region then would it make sense to fail fast?

awsutil/awsutil_test.go
```diff
@@ -0,0 +1,76 @@
...
+		},
+		{
+			name: "aws-region-and-default-region",
+			env: map[string]string{
+				"AWS_REGION":          nonTaskRegion,
+				"AWS_DEFAULT_REGION":  "should-not-use-this-one",
+				"AWS_SDK_LOAD_CONFIG": "1",
+			},
+			expectRegion: nonTaskRegion,
+		},
```
ishustava, line 54: If the logic for the different region env vars lives in the AWS SDK, what is the reason for having the tests for it here? It seems like the only way to make them fail is to not call `session.NewSession()`, but then you wouldn't be creating an api client and the test would fail since the session would be nil. 

If we don't have any logic related to reading the different env variables, then we should remove tests for that since I think they're testing the behavior of the SDK rather than our own logic.

awsutil/awsutil_test.go
```diff
@@ -0,0 +1,76 @@
...
+	}
+
+	for _, c := range cases {
+		t.Run(c.name, func(t *testing.T) {
+			os.Clearenv()
```
ishustava, line 59: would that clear all env variables even the ones we don't set? should we instead call `defer` to the env vars we're setting? If we would want to set env vars for tests at some point in the future, that could cause weird bugs that will take a while to debug.

ishustava (approved):
Great work! I left some comments, but don't think they're blocking.

awsutil/awsutil_test.go
pglass: That's fair. Though, it was only after writing these tests that I felt confident that we respect the `AWS_REGION`/`AWS_DEFAULT_REGION` environment variables if they are set... 

​While our code doesn't explicitly read those environment variables, it does configure the session and there are certainly ways to configure the session which would cause those those environment variables to be ignored (i.e. assume I'll write bad code in the future. Tests will save me.)

At minimum, I'd want to keep the first two test cases: (1) validate we _do_ fallback to the Task metadata with an empty env, and (2) validate we _don't_ cause `AWS_REGION` to be ignored if it's set.

awsutil/awsutil_test.go
pglass: Yeah, this is a test isolation concern. Let me fix this.

awsutil/awsutil_test.go
ishustava: Agree on `#1`. For the `#2`, if there is a way we can write the code to ignore `AWS_REGION`, then it makes sense to test it. Ultimately, I'm advocating for tests that can fail and therefore can provide us feedback in the future. I couldn't come up with code that would make the `aws-region-and-default-region` test fail, for example, but maybe I just don't have vivid imagination 😄 

Either way, it's not blocking and so I'm OK with either decision!

pglass: So for (2), if I accidentally invert a conditional in `awsutil.NewSession`:

```
        cfg := clientSession.Config
	if cfg.Region == nil || *cfg.Region != "" {  // bug!
		cfg.Region = aws.String(meta.Region())
	}
```

That results in us overwriting the environment-discovered region, despite not explicitly reading any of those variables. So including at least one test that sets AWS_REGION is useful, since it helps validate our implementation. 

But I agree we don't need multiple such tests with various combinations of environment variables.

awsutil/awsutil_test.go
pglass: Okay, I addressed this in the newer with a `restoreEnv` function and defers (`t.Cleanup`, actually). I'm not sure I like it.

If we did drop all the tests of environment variables, we could remove this env cleanup stuff as well, and I'd be fine with that too.

ishustava: Oh I see. It's because somebody might have it in their env before running the test and we want to make sure it doesn't affect us and doesn't mutate someone's environment. 🤔 yeah, I'm not sure if there's a cleaner way than what you did. Generally, testing with env vars is kind of messy in my experience, so I'm OK with what you wrote.

PR #22, opened 2021-08-31
Implement ACL controller

ishustava:
* Add a new command called "acl-controller." The command will first
  create the token for the Consul client and then will start a controller
  to manage service tokens.
* The controller for the service tokens polls for tasks periodically (10s)
  and creates tokens for any tasks it hasn't seen (reconciliation).
* Currently, controller only handles upserts but not deletes. Deletes
  will be implemented later.

Acceptance tests in hashicorp/terraform-aws-consul-ecs#31

Note for the reviewers:
I did not add any unit tests for the command itself for two reasons:
1. Most of the logic is tested in the `controller` package.
2. The rest of it (such as parsing the CA cert  for example) is tested by acceptance tests in hashicorp/terraform-aws-consul-ecs.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,0 +1,5 @@
+issues:
+  exclude-rules:
+    - path: _test\.go
+      linters:
+        - errcheck
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,5 +1,0 @@
-issues:
-  exclude-rules:
-    - path: _test\.go
-      linters:
-        - errcheck
--- a/controller/consul_client_token.go
+++ b/controller/consul_client_token.go
@@ -39,55 +39,60 @@ func UpsertConsulClientToken(consulClient *api.Client, smClient secretsmanagerif
 		}
 	}
 
-	// If there is no current token in Consul, we need to create one.
-	if currToken == nil {
-		// First, we need to check if the policy for the Consul client already exists.
-		// If it does, we will skip policy creation.
-		policyName := fmt.Sprintf("%s-consul-client-policy", secretPrefix)
-		policy, _, err := consulClient.ACL().PolicyReadByName(policyName, nil)
+	// Exit if current token is found in Consul.
+	if currToken != nil {
+		return nil
+	}
+	// Otherwise, we need to create one.
+	// First, we need to check if the policy for the Consul client already exists.
+	// If it does, we will skip policy creation.
+	policyName := fmt.Sprintf("%s-consul-client-policy", secretPrefix)
+	policy, _, err := consulClient.ACL().PolicyReadByName(policyName, nil)
 
-		// When policy is not found, Consul returns ACL not found error.
-		if isACLNotFoundError(err) {
-			// Create a policy for the Consul clients.
-			log.Info("creating ACL policy", "name", policyName)
-			policy, _, err = consulClient.ACL().PolicyCreate(&api.ACLPolicy{
-				Name:        policyName,
-				Description: "Consul Client Token Policy for ECS",
-				Rules:       `node_prefix "" { policy = "write" } service_prefix "" { policy = "read" }`,
-			}, nil)
-			if err != nil {
-				return fmt.Errorf("creating Consul client ACL policy: %w", err)
-			}
-			log.Info("ACL policy created successfully", "name", policyName)
-		} else if err != nil {
-			return fmt.Errorf("reading Consul client ACL policy: %w", err)
-		}
-
-		log.Info("creating Consul client ACL token")
-		token, _, err := consulClient.ACL().TokenCreate(&api.ACLToken{
-			Description: "ECS Consul client Token",
-			Policies:    []*api.ACLTokenPolicyLink{{Name: policy.Name}},
+	// When policy is not found, Consul returns ACL not found error.
+	if isACLNotFoundError(err) {
+		// Create a policy for the Consul clients.
+		log.Info("creating ACL policy", "name", policyName)
+		policy, _, err = consulClient.ACL().PolicyCreate(&api.ACLPolicy{
+			Name:        policyName,
+			Description: "Consul Client Token Policy for ECS",
+			// We use this token for all Consul clients, and that's why the node_prefix needs to be "".
+			Rules: `node_prefix "" { policy = "write" } service_prefix "" { policy = "read" }`,
 		}, nil)
 		if err != nil {
-			return fmt.Errorf("creating Consul client ACL token: %w", err)
+			return fmt.Errorf("creating Consul client ACL policy: %w", err)
 		}
-		log.Info("Consul client ACL token created successfully")
-
-		agentSecret, err := json.Marshal(tokenSecretJSON{Token: token.SecretID, AccessorID: token.AccessorID})
-		if err != nil {
-			return fmt.Errorf("marshalling Consul client token: %w", err)
-		}
-
-		// Finally, update the AWS Secret with the new values of the token.
-		log.Info("updating secret", "arn", secretARN)
-		_, err = smClient.UpdateSecret(&secretsmanager.UpdateSecretInput{
-			SecretId:     aws.String(secretARN),
-			SecretString: aws.String(string(agentSecret)),
-		})
-		if err != nil {
-			return fmt.Errorf("updating secret: %s", err)
-		}
-		log.Info("secret updated successfully", "arn", secretARN)
+		log.Info("ACL policy created successfully", "name", policyName)
+	} else if err != nil {
+		return fmt.Errorf("reading Consul client ACL policy: %w", err)
+	} else {
+		log.Info("ACL policy already exists; skipping policy creation", "name", policyName)
 	}
+
+	log.Info("creating Consul client ACL token")
+	token, _, err := consulClient.ACL().TokenCreate(&api.ACLToken{
+		Description: "ECS Consul client Token",
+		Policies:    []*api.ACLTokenPolicyLink{{Name: policy.Name}},
+	}, nil)
+	if err != nil {
+		return fmt.Errorf("creating Consul client ACL token: %w", err)
+	}
+	log.Info("Consul client ACL token created successfully")
+
+	clientSecret, err := json.Marshal(tokenSecretJSON{Token: token.SecretID, AccessorID: token.AccessorID})
+	if err != nil {
+		return fmt.Errorf("marshalling Consul client token: %w", err)
+	}
+
+	// Finally, update the AWS Secret with the new values of the token.
+	log.Info("updating secret", "arn", secretARN)
+	_, err = smClient.UpdateSecret(&secretsmanager.UpdateSecretInput{
+		SecretId:     aws.String(secretARN),
+		SecretString: aws.String(string(clientSecret)),
+	})
+	if err != nil {
+		return fmt.Errorf("updating secret: %s", err)
+	}
+	log.Info("secret updated successfully", "arn", secretARN)
 	return nil
 }
--- a/controller/consul_client_token.go
+++ b/controller/consul_client_token.go
@@ -1,0 +1,93 @@
+package controller
+
+import (
+	"encoding/json"
+	"fmt"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/secretsmanager"
+	"github.com/aws/aws-sdk-go/service/secretsmanager/secretsmanageriface"
+	"github.com/hashicorp/consul/api"
+	"github.com/hashicorp/go-hclog"
+)
+
+// UpsertConsulClientToken creates or updates ACL policy and token for the Consul client in Consul.
+// It then saves the created token in AWS Secrets Manager in the secret provided by secretARN.
+func UpsertConsulClientToken(consulClient *api.Client, smClient secretsmanageriface.SecretsManagerAPI, secretARN, secretPrefix string, log hclog.Logger) error {
+	// Read the secret from AWS.
+	currSecretValue, err := smClient.GetSecretValue(&secretsmanager.GetSecretValueInput{
+		SecretId: aws.String(secretARN),
+	})
+	if err != nil {
+		return fmt.Errorf("retrieving Consul client secret: %w", err)
+	}
+
+	// Unmarshal the secret value JSON.
+	var currSecret tokenSecretJSON
+	err = json.Unmarshal([]byte(*currSecretValue.SecretString), &currSecret)
+	if err != nil {
+		return fmt.Errorf("unmarshalling Consul client secret value JSON: %w", err)
+	}
+
+	var currToken *api.ACLToken
+
+	// If the secret is not empty, check if Consul already has a token with this AccessorID.
+	if currSecret.AccessorID != "" {
+		currToken, _, err = consulClient.ACL().TokenRead(currSecret.AccessorID, nil)
+		if err != nil && !isACLNotFoundError(err) {
+			return fmt.Errorf("reading token: %w", err)
+		}
+	}
+
+	// If there is no current token in Consul, we need to create one.
+	if currToken == nil {
+		// First, we need to check if the policy for the Consul client already exists.
+		// If it does, we will skip policy creation.
+		policyName := fmt.Sprintf("%s-consul-client-policy", secretPrefix)
+		policy, _, err := consulClient.ACL().PolicyReadByName(policyName, nil)
+
+		// When policy is not found, Consul returns ACL not found error.
+		if isACLNotFoundError(err) {
+			// Create a policy for the Consul clients.
+			log.Info("creating ACL policy", "name", policyName)
+			policy, _, err = consulClient.ACL().PolicyCreate(&api.ACLPolicy{
+				Name:        policyName,
+				Description: "Consul Client Token Policy for ECS",
+				Rules:       `node_prefix "" { policy = "write" } service_prefix "" { policy = "read" }`,
+			}, nil)
+			if err != nil {
+				return fmt.Errorf("creating Consul client ACL policy: %w", err)
+			}
+			log.Info("ACL policy created successfully", "name", policyName)
+		} else if err != nil {
+			return fmt.Errorf("reading Consul client ACL policy: %w", err)
+		}
+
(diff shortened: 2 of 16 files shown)


controller/resource.go
```diff
@@ -0,0 +1,222 @@
...
+
+	Cluster      string
+	SecretPrefix string
+	Task         *ecs.Task
+
+	Log hclog.Logger
+}
+
+// ID returns Task family/Consul service name or error if it cannot be determined from the Task group.
+// It is derived from the group which looks like "service:<service name>".
```
pglass, line 107: Hmm, non-blocking: The [docs](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task-groups.html) mention the Task group could be entirely custom (which seems uncommon). Would parsing the family from the `TaskDefinitionArn` be more reliable, since it's a known format? (I don't see another field in the Task that includes the family ☹️ )

controller/resource.go
```diff
@@ -0,0 +1,222 @@
...
+			})
+		}
+		if nextToken == nil {
+			break
+		}
+	}
+	return resources, nil
+}
+
+type Tasks struct {
```
pglass, line 95: naming nit: `Tasks` had me expecting multiple tasks involved. This represents the Task family, so maybe `Task` or `TaskFamily`. Thoughts?

pglass (approved):
Overall, looks great after a first pass! 👍

.golangci.yml
```diff
@@ -0,0 +1,5 @@
+issues:
+  exclude-rules:
+    - path: _test\.go
+      linters:
+        - errcheck
```
lkysow, line 5: should we still have it to ensure the tests aren't ignoring errors?

subcommand/controller/command.go
```diff
@@ -0,0 +1,117 @@
...
+
+	err := c.run()
+	if err != nil {
+		c.log.Error(err.Error())
+		return 1
+	}
+	return 0
+}
+
+func (c *Command) run() error {
```
lkysow, line 57: I think the pattern I've seen is using `realRun` or something like that. (I didn't ever know Go's functions were case sensitive!)

subcommand/controller/command.go
```diff
@@ -0,0 +1,117 @@
...
+
+func (c *Command) Run(args []string) int {
+	c.once.Do(c.init)
+	if err := c.flagSet.Parse(args); err != nil {
+		return 1
+	}
+
+	err := c.run()
+	if err != nil {
+		c.log.Error(err.Error())
```
lkysow, line 51: 🙏 banish `c.UI` 😂

subcommand/controller/command.go
```diff
@@ -0,0 +1,117 @@
...
+		SecretPrefix:         c.flagSecretNamePrefix,
+		Log:                  c.log,
+	}
+	ctrl := controller.Controller{
+		Resources:       taskTokens,
+		PollingInterval: controller.DefaultPollingInterval,
+		Log:             c.log,
+	}
+
+	ctrl.Run(context.Background())
```
lkysow, line 106: does this just block?

subcommand/controller/command.go
```diff
@@ -0,0 +1,117 @@
...
+	"github.com/hashicorp/consul/api"
+	"github.com/hashicorp/go-hclog"
+	"github.com/mitchellh/cli"
+)
+
+const (
+	flagAgentSecretARN   = "agent-secret-arn"
+	flagSecretNamePrefix = "secret-name-prefix"
+
+	consulCACertEnvVar = "CONSUL_CACERT_PEM"
```
lkysow, line 22: is it a pem because we're injecting via env var and you don't want to write it to file?

subcommand/controller/command.go
```diff
@@ -0,0 +1,117 @@
...
+	if err != nil {
+		return err
+	}
+
+	smClient := secretsmanager.New(clientSession, nil)
+
+	err = controller.UpsertConsulClientToken(consulClient, smClient, c.flagAgentSecretARN, c.flagSecretNamePrefix, c.log)
+	if err != nil {
+		return err
+	}
```
lkysow, line 90: Should we straight up error here? Can the controller loop instead?

controller/consul_client_token.go
```diff
@@ -0,0 +1,93 @@
...
+		policy, _, err := consulClient.ACL().PolicyReadByName(policyName, nil)
+
+		// When policy is not found, Consul returns ACL not found error.
+		if isACLNotFoundError(err) {
+			// Create a policy for the Consul clients.
+			log.Info("creating ACL policy", "name", policyName)
+			policy, _, err = consulClient.ACL().PolicyCreate(&api.ACLPolicy{
+				Name:        policyName,
+				Description: "Consul Client Token Policy for ECS",
+				Rules:       `node_prefix "" { policy = "write" } service_prefix "" { policy = "read" }`,
```
lkysow, line 56: worth a note that all clients share a token so that's why we're using prefix empty?

controller/consul_client_token.go
```diff
@@ -0,0 +1,93 @@
...
+			}, nil)
+			if err != nil {
+				return fmt.Errorf("creating Consul client ACL policy: %w", err)
+			}
+			log.Info("ACL policy created successfully", "name", policyName)
+		} else if err != nil {
+			return fmt.Errorf("reading Consul client ACL policy: %w", err)
+		}
```
lkysow, line 64: worth a log that the policy already existed?

controller/consul_client_token.go
```diff
@@ -0,0 +1,93 @@
...
+		token, _, err := consulClient.ACL().TokenCreate(&api.ACLToken{
+			Description: "ECS Consul client Token",
+			Policies:    []*api.ACLTokenPolicyLink{{Name: policy.Name}},
+		}, nil)
+		if err != nil {
+			return fmt.Errorf("creating Consul client ACL token: %w", err)
+		}
+		log.Info("Consul client ACL token created successfully")
+
+		agentSecret, err := json.Marshal(tokenSecretJSON{Token: token.SecretID, AccessorID: token.AccessorID})
```
lkysow, line 76: we use agent here but client elsewhere. Should we standardize on "client"?

controller/consul_client_token.go
```diff
@@ -0,0 +1,93 @@
...
+	// If the secret is not empty, check if Consul already has a token with this AccessorID.
+	if currSecret.AccessorID != "" {
+		currToken, _, err = consulClient.ACL().TokenRead(currSecret.AccessorID, nil)
+		if err != nil && !isACLNotFoundError(err) {
+			return fmt.Errorf("reading token: %w", err)
+		}
+	}
+
+	// If there is no current token in Consul, we need to create one.
+	if currToken == nil {
```
lkysow, line 43: Can we do `if currToken != nil { return nil }` and exit early?

lkysow (commented):
I'm only half way through but wanted to submit what I've got so far. Looks really good.

controller/resource.go
ishustava: That's a good point. And actually, from the docs it seems that if you have a service, then the group name must be the service name. In our case, the service name is not guaranteed to be equal to the task family name. I'll change it to parse it from the task def ARN, thanks for doing the research!!

ishustava: 🤔 Thinking about this more, I'm thinking to switch to tracking state with the whole Task Definition ARN instead of task family. The reason for that is that task def is mutable, and so if you, for example, update your task from not being on the mesh to now use the mesh-task, we won't detect that since we'll think that we've already "seen" this task family and won't know that it now has the mesh tag.

Because the task definition ARN will also include the revision version, we can therefore detect that the task definition has changed from the one we "know" about and should process it again.

controller/resource.go
ishustava: Yep, good point. Who named these 😄 

I'll change it to `Task` since it wraps `ecs.Task` type.

.golangci.yml
ishustava: Yeah, I can add it back in.

subcommand/controller/command.go
ishustava: Oh sorry, I didn't realize it was a pattern! I changed it because personally I found it a bit confusing because it made me think that the `Run()` method is not "real" 😸.  Where does the `realRun` pattern come from? I don't think I've seen it before. 

Yeah, golang is case-sensitive. That's how it decides if a function is exported or not. I've seen `run()` method in the Consul codebase being used a few times, e.g. [here](https://github.com/hashicorp/consul/blob/a2e6a15e936b4907bf6a1790d91da96ee8949997/command/connect/envoy/envoy.go#L229), but don't think I can call it a pattern.

subcommand/controller/command.go
ishustava: That's from your POC I think 😄 and I completely agree!

subcommand/controller/command.go
ishustava: yep, it blocks. There are no other processes though, and so we don't need to run it in a go func and then have a for/select loop to wait for signals I think.

subcommand/controller/command.go
ishustava: yeah exactly. Also, if you try to re-use the `CONSUL_CACERT` env var which Consul thinks points to a file, everything breaks even you overwrite values returned by `DefaultConfig()` because `api.NewClient` calls `defaultConfig` again (speaking from experience and a lot of printf debugging 😄 ).

controller/resource.go
pglass: True!! Basically, if any version of a TaskDefinition is tagged as a mesh task, the controller should ensure an ACL token is provisioned for that TaskDefinition family.

Is that right? We could optimize slightly based on that: once an ACL token is provisioned for the TaskDefinition family, we don't need to examine any other versions of that TaskDefinition (non-blocking, for sure)

subcommand/controller/command.go
ishustava: This isn't currently called by controller (it just lives in the controller package), instead right now it's a stand-alone function that's called during start up of the process. 

We can make it part of the controller loop, although I'd prefer to make that change in a future PR to keep the scope smaller.

subcommand/controller/command.go
lkysow: I was taking it from TF (https://github.com/hashicorp/terraform/blob/main/main.go#L61). But if Consul already has their own pattern then it makes more sense to use theirs.

controller/resource.go
ishustava: > once an ACL token is provisioned for the TaskDefinition family, we don't need to examine any other versions of that TaskDefinition

Yeah, good idea! I think that's worth investigating.

subcommand/controller/command.go
lkysow: What if the command in the container first wrote it to file before executing `consul-ecs acl-controller`?

subcommand/controller/command.go
lkysow: What about if it fails though, for example the consul server isn't up yet.

ishustava: If the consul server isn't up, then the task will be restarted. I think that's OK since in most cases (except for the dev-server), the server should already be up.

subcommand/controller/command.go
ishustava: It's also doable, although it would require us to overwrite the entry point of the container which we currently are not doing. I didn't want to mess with that. I could see though that there advantages and disadvantages to either solution.

controller/controller.go
```diff
@@ -0,0 +1,70 @@
...
+	for _, resource := range resources {
+		resourceID, err := resource.ID()
+		if err != nil {
+			return fmt.Errorf("getting resource ID: %w", err)
+		}
+		if _, ok := c.resourceState[resourceID]; !ok {
+			err = resource.Upsert()
+			if err != nil {
+				return fmt.Errorf("upserting resource: %w", err)
+			}
```
lkysow, line 63: Does exiting here have an issue where we loop on the same erroring item over and over and don't process the other items in the list?

controller/controller.go
```diff
@@ -0,0 +1,70 @@
...
+	// Resources lists resources for Controller to reconcile.
+	Resources ResourceLister
+	// PollingInterval is an interval that Controller will use to reconcile all Resources.
+	PollingInterval time.Duration
+	// Log is the logger used by the Controller.
+	Log hclog.Logger
+
+	// resourceState is the internal Controller's state. The Controller
+	// will periodically update the state to reflect the state of the Resources.
+	resourceState map[string]struct{}
```
lkysow, line 26: This is a map of the ids of the resources that we currently know to exist and that have successfully been upserted once right? Should we state that more precisely in the comment or are you leaving this a bit more vague as "internal state" because it might be used for something else later?

controller/resource.go
```diff
@@ -0,0 +1,240 @@
...
+		tasks, err := t.ECSClient.DescribeTasks(&ecs.DescribeTasksInput{
+			Cluster: aws.String(t.Cluster),
+			Tasks:   taskListOutput.TaskArns,
+			Include: []*string{aws.String("TAGS")},
+		})
+		if err != nil {
+			return nil, fmt.Errorf("describing tasks: %w", err)
+		}
+		for _, task := range tasks.Tasks {
+			resources = append(resources, &Task{
```
lkysow, line 79: should we check here if `task` is nil and if it is error and if it's not dereference it so we don't have to worry about nil pointers in the `Task` struct?

controller/resource.go
```diff
@@ -0,0 +1,240 @@
...
+type Task struct {
+	SecretsManagerClient secretsmanageriface.SecretsManagerAPI
+	ConsulClient         *api.Client
+
+	Cluster      string
+	SecretPrefix string
+	Task         *ecs.Task
+
+	Log hclog.Logger
+}
```
lkysow, line 104: This set of fields kinda confuses me to be on `Task`. I see we need them because we're caling `task.Upsert()` but I'm curious why we didn't go with an interface of `TaskController.Upsert(*ecs.Task)` and that `TaskController` struct would have all the `SecretsManagerClient`, `ConsulClient`, etc fields.

For example our config entry ctrls in k8s have this interface where they take in a `req` which is essentially the `task`:

```go
func (r *ConfigEntryController) ReconcileEntry(ctx context.Context, crdCtrl Controller, req ctrl.Request, configEntry common.ConfigEntryResource)
```

lkysow: So then I guess `controller` would be more the ControllerManager in my head and then we'd write an `ACLController` that has interfaces:

```go
Upsert(*ecs.Task)
Delete(*ecs.Task)
```

controller/resource.go
```diff
@@ -0,0 +1,240 @@
...
+		// Read the token with this Accessor ID from Consul.
+		currToken, _, err = t.ConsulClient.ACL().TokenRead(currSecret.AccessorID, nil)
+
+		if err != nil && !isACLNotFoundError(err) {
+			return fmt.Errorf("reading existing token: %w", err)
+		}
+	}
+
+	// If there is already a token for this service in Consul, exit early.
+	if currToken != nil {
```
lkysow, line 165: Should we log something for this case?

lkysow: I guess you've only got error logs here. Should we have some info logs letting users know that a token was creatd?

lkysow (approved):
This looks great. I've got some questions about the interfaces but they shouldn't block our beta release.

subcommand/controller/command.go
lkysow: I think I like this controller just running until it works rather than erroring but don't feel too strongly about it.

subcommand/controller/command.go
lkysow: I think for consistency keeping `CONSUL_CACERT` across things would be better but don't feel too strongly.

subcommand/controller/command.go
lkysow: makes sense!

controller/controller.go
pglass: In addition to the comment, we could add a type def to clarify this too, e.g.

```go
type ResourceId string

type Resource interface {
	ID() (ResourceId, error)
        ...
}

type Controller struct {
        resourceState map[ResourceId]struct{}
        ...
}
```

Totally optional.

ishustava: >This is a map of the ids of the resources that we currently know to exist and that have successfully been upserted once right

Right.

>In addition to the comment, we could add a type def to clarify this too, e.g.

Yeah, I like that idea!

controller/controller.go
ishustava: you're right, great catch! I think we need to keep processing other resources in the loop rather than exiting here.

controller/resource.go
ishustava: Yeah, I see what you mean about the fields. The idea here was to have a generic controller implementation that can reconcile any kind of resource. Then it'd be up to the resource's implementation to decide what it means for it to upsert or to delete itself. For that, it needs to have external store parameters, like the clients etc so that it can do those things. This approach was probably more inspired by the [`ServiceResource`](https://github.com/hashicorp/consul-k8s/blob/c79730ba4f0e25b8330950d7ab575e4e3cfc1ef0/control-plane/catalog/to-consul/resource.go#L51) we have in the k8s sync-catalog rather than the `controller-runtime` type controllers. `ServiceResource` similarly is a type that has fields it needs to "reconcile" itself. The reason I went with this approach is that it seemed more intuitive to me for this use case. Meaning it seems more intuitive to say `resource.Upsert` rather than `controller.Upsert`, but this might be just a personal preference.

I think what you're suggesting could also work. Other than what I said above, I don't have a super-strong opinion on which way is better. If we do want to change it, I'd do it in a separate PR.

subcommand/controller/command.go
```diff
@@ -0,0 +1,117 @@
...
+		Log:             c.log,
+	}
+
+	ctrl.Run(context.Background())
+
+	return nil
+}
+
+func (c *Command) Synopsis() string {
+	return "ECS controller"
```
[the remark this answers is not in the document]
ishustava: I think it should be called ACL controller

controller/resource.go
lkysow: Agreed on separate PR. I prefer the Kubernetes controller approach but I think that's also just a personal preference thing. @pglass something to think about after this release.

PR #29, opened 2021-09-20
Support Consul native health checks

pglass:
Add `-checks` flag to mesh-init to support passing in Consul native checks


subcommand/mesh-init/command_test.go
```diff
@@ -148,6 +186,22 @@ func TestRun(t *testing.T) {
...
+				for _, expCheck := range c.checks {
+					require.Contains(t, checks, expCheck.CheckID)
+				}
+
+				// Ensure health status is "critical".
+				// These tests do not start a listening application, so the checks will not pass.
+				healthStatus, checksInfo, err := consulClient.Agent().AgentHealthServiceByID(service.ID)
+				require.NoError(t, err)
+				require.NotNil(t, checksInfo)
+				require.Equal(t, healthStatus, api.HealthCritical)
```
lkysow, line 202: nice work testing this too!

subcommand/mesh-init/command_test.go
```diff
@@ -148,6 +186,22 @@ func TestRun(t *testing.T) {
 			require.NoError(t, err)
 			require.True(t, cmp.Equal(expectedProxyServiceRegistration, proxyService, agentServiceIgnoreFields))
 
+			if c.checks != nil {
+				// Validate the check is present
+				checks, err := consulClient.Agent().Checks()
+				require.NoError(t, err)
+				for _, expCheck := range c.checks {
+					require.Contains(t, checks, expCheck.CheckID)
```
lkysow, line 194: Is there a way to assert against the check definition to make sure the fields got set?

lkysow (commented):
Looks good. One question about the tests.

subcommand/mesh-init/command_test.go
```diff
@@ -55,6 +59,37 @@ func TestRun(t *testing.T) {
 				},
 			},
 		},
+		"service with checks": {
+			checks: api.AgentServiceChecks{
+				&api.AgentServiceCheck{
+					// check id should be "api-<type>" for assertions
```
ishustava, line 65: Minor edit 😃 
```suggestion
					// Check id should be "api-<type>" for assertions.
```

subcommand/mesh-init/command_test.go
```diff
@@ -151,6 +191,51 @@ func TestRun(t *testing.T) {
...
 			require.NotEmpty(t, envoyBootstrapContents)
+
+			if c.checks != nil {
+				actualChecks, err := consulClient.Agent().Checks()
+				require.NoError(t, err)
+				for _, expCheck := range c.checks {
+					expectedAgentCheck := toAgentCheck(expCheck)
+					// Check for "critical" status. There is no listening application here, so checks will not pass.
+					expectedAgentCheck.Status = api.HealthCritical
+					// Pull the check type from the CheckID: "api-<type>" -> "<type>"
```
ishustava, line 202: nit: Might be worth double-checking other comments too.
```suggestion
					// Pull the check type from the CheckID: "api-<type>" -> "<type>".
```

subcommand/mesh-init/command_test.go
```diff
@@ -151,6 +191,51 @@ func TestRun(t *testing.T) {
...
+					expectedAgentCheck.Type = strings.ReplaceAll(expCheck.CheckID, "api-", "")
```
ishustava, line 203: Why do we need to do that?

ishustava (approved):
Looks good! I haven't tested it myself though.

subcommand/mesh-init/command_test.go
lkysow, line 202:
```suggestion
					// Pull the check type from the CheckID: "api-<type>" -> "<type>"
					// because Consul adds the Type field in its response.
```

lkysow (approved):

subcommand/mesh-init/command_test.go
pglass: Consul has a `type` field in the response, but not the request ([link to request examples](https://www.consul.io/docs/discovery/checks#check-definition)). 

So I encoded type into the check id. 

I guess I could make the check id be the exact check type to simplify. Or, I could include the expected check type as a separate field in the test case data.

subcommand/mesh-init/command_test.go
pglass: (I used Luke's suggestion, sorry!)

subcommand/mesh-init/command_test.go
pglass: I'm going to merge as-is - but happy to follow up and change/simplify this if you think we should! 😀

PR #30, opened 2021-09-22
Delete ACL tokens for task families that are deleted.

ishustava:
## Changes proposed
- ACL controller's reconcile now also reconciles its internal state with existing resources
- For any resource that is in the internal state but doesn't exist, it calls `resource.Delete`
- Acceptance tests are in hashicorp/terraform-aws-consul-ecs#40

## Notes
I have decided to keep everything with the same loop (as opposed to having a separate delete loop with a different interval) to avoid having a separate state that tracks current resources. I felt that it made the logic simpler, and I don't think it'll add a lot of extra overhead in terms of AWS API calls. Let me know if you think differently.


controller/resource.go
```diff
@@ -174,17 +173,62 @@ func (t *Task) Upsert() error {
...
+				SecretString: aws.String(`{}`),
+			})
+			if err != nil {
+				return fmt.Errorf("updating secret: %s", err)
+			}
+			t.Log.Info("secret updated successfully", "name", secretName)
+		}
+	}
+	if !tokenFound {
+		return fmt.Errorf("could not find token for service %q", serviceName)
```
lkysow, line 230: Should this return an error? Doesn't this mean that we're 1/2 way through deleting and we should double check that the SM secret is also zero'd out?

For example, what if our consul token delete succeeds but then the SM update fails. Then we'll be in an infinite loop of trying to delete the token.

controller/resource.go
```diff
@@ -174,17 +173,62 @@ func (t *Task) Upsert() error {
...
+			secretName := t.secretName(serviceName)
+			t.Log.Info("updating secret", "name", secretName)
+			_, err = t.SecretsManagerClient.UpdateSecret(&secretsmanager.UpdateSecretInput{
+				SecretId:     aws.String(secretName),
+				SecretString: aws.String(`{}`),
+			})
+			if err != nil {
+				return fmt.Errorf("updating secret: %s", err)
+			}
+			t.Log.Info("secret updated successfully", "name", secretName)
```
lkysow, line 226: can we have the service/servicename key/value in this log too?

controller/resource.go
```diff
@@ -199,6 +243,7 @@ func (t *Task) updateServiceToken(serviceName, secretName string) error {
 	if err != nil {
 		return fmt.Errorf("creating envoy token: %s", err)
 	}
+	t.Log.Info("service token created successfully", "id", serviceName)
```
lkysow, line 246:
```suggestion
	t.Log.Info("service token created successfully", "service", serviceName)
```

For consistency. id has been used as the taskARN elsewhere and service was used for serviceName above

lkysow (commented):
I still find it confusing that the resource ID is the taskdefinition but the resource is the task. What about renaming `Task` => `TaskDef`? Then `TaskDef.Upsert()` and `TaskDef.Delete()` make more sense to me.

(Or maybe I just need to have more time with this setup)

controller/controller.go
```diff
@@ -65,10 +69,25 @@ func (c *Controller) reconcile() error {
...
+	}
+
+	for id, resource := range c.resourceState {
+		if _, ok := currentResourceIDs[id]; !ok {
+			// Delete the resource.
+			err = resource.Delete()
+			if err != nil {
+				// If there's an error, log it but don't delete from internal
+				// state so that we can retry on the next iteration of the 'reconcile' loop.
+				c.Log.Error("error deleting resource", "err", err)
```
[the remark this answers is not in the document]
ishustava: yeah, that's a good call-out.

controller/resource.go
ishustava: 🤔 this if statement is outside the loop. We will return an error if we looked through all tokens and haven't found one for the service. Not sure what you mean.

controller/resource.go
```diff
@@ -219,7 +261,8 @@ func (t *Task) updateServiceToken(serviceName, secretName string) error {
 }
 
 // Task definition ARN looks like this: arn:aws:ecs:us-east-1:1234567890:task-definition/service:1
-func (t *Task) parseFamilyNameFromTaskDefinitionARN(taskDefArn string) (string, error) {
+func (t *Task) parseFamilyNameFromTaskDefinitionARN() (string, error) {
+	taskDefArn := *t.Task.TaskDefinitionArn
```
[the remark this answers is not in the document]
ishustava: yeah, it's not a pointer, but the `TaskDefinitionArn` inside `ecs.Task` is. That's what we're de-referencing here.

ishustava:
> I still find it confusing that the resource ID is the taskdefinition but the resource is the task. What about renaming Task => TaskDef? Then TaskDef.Upsert() and TaskDef.Delete() make more sense to me.

Yeah, it's a bit inconsistent right now because we only use task def for the ID but tasks for everything else. Meaning that the `TaskLister` lists `ecs.Task`s and also the `Task` struct wraps `ecs.Task`. So if we rename those to be `TaskDefinition` then it'll be confusing to see that the implementation is to list tasks in instead of task definitions, for example. Maybe we should just not worry about the lack of batch APIs and switch everything to be task defs?

controller/resource.go
lkysow: I created https://github.com/hashicorp/consul-ecs/pull/31 to better illustrate. It's a bit of an edge case but I'm asking what happens if the `UpdateSecret` call fails for some reason.

lkysow:
I think listing tasks as a proxy for finding task definitions for tasks that are running is totally fine. If we comment in the List() function that "there's no batch API for finding task definitions so we're using listtasks and then iterating through those" that would make sense to me.

I'm also okay leaving it as is if it's not an issue for others.

controller/resource.go
ishustava: Ooh, great catch!! Thanks so much for creating the code example too, that was super helpful ❤️ 

I think you're right that this error doesn't make sense and that we should instead make behavior idempotent, in that if there's no token, we should still update existing secret. I've updated the code and added test for this use case.

controller/resource.go
```diff
@@ -110,6 +110,9 @@ type Task struct {
 }
 
 // ID returns Task definition ARN or error if it cannot be determined from the Task.
+// We're using task definition ARN instead of the Task ARN because we consider tasks
+// for the same definition to be equivalent in terms of their Upsert and Delete behavior,
+// and so there's no reason to process subsequent tasks with the same task definition.
```
ishustava, line 115: @lkysow I've added clarifying comment here instead. 

I definitely see what you mean though. On the one hand, I'd expect the Task's ID to be task ARN. On the other, I think a task could have a different ID depending on what makes sense for your implementation, and in this case, we can think of it as tasks with the same definition are equal to each other.

I'm a bit hesitant about the rename you suggested because I find it a bit more confusing than this for two reasons:
1. If rename `TaskLister` to `TaskDefinitionLister` it's be inconsistent in that it's actually listing tasks. I know you've mentioned that we can help clarify that with a comment, but that's where the second reason comes in.
2. The `Task` resource is a wrapper around `ecs.Task` struct, and so if we rename it to be `TaskDefinition` instead but it'll still have a reference to `ecs.Task`, I think it'll also be confusing.

Also, happy to hear feedback from @pglass to see what makes more sense. I could have been spending too much time with this code and just have tunnel vision.

controller/resource.go
lkysow: Sorry, I don't see that test case yet?

ishustava: Sorry my comment was confusing. I ended up modifying the existing test case when a token doesn't exist in consul but I realized after looking at it now that I need to update it to test the case when token doesn't exist but the secret has some value.

ishustava: Updated. I've split into different scenarios covering permutations of either token existing or secret having some value in it: 

https://github.com/hashicorp/consul-ecs/blob/ca1b6473f74f891207b8a893fe5a5bbafb11553f/controller/resource_test.go#L196-L235

lkysow (approved):

controller/controller.go
```diff
@@ -46,29 +48,48 @@ func (c *Controller) Run(ctx context.Context) {
...
-			c.resourceState[resourceID] = struct{}{}
+			c.resourceState[resourceID] = resource
+		}
+	}
+
+	// Prune any resources in the resource state that no longer exist.
+	for id, resource := range c.resourceState {
+		if _, ok := currentResourceIDs[id]; !ok {
+			// Delete the resource.
+			err = resource.Delete()
```
pglass, line 81: Trying to understand when we actually call `resource.Delete()`:

* `c.Resources.List()` returns a list of Tasks (effectively)
* `c.resourceState` is keyed by the TaskDefinitionARN
* ListTasks returns recently stopped tasks [for at least one hour](https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_ListTasks.html)

So, @ishustava is this a problematic sequence of events, or am I misunderstanding?

* I stop all tasks for TaskDefinitionARN `my-task:1`
* I start tasks for TaskDefinitionARN `my-task:2`
* The controller calls `resource.Upsert()` for `my-task:2`, does nothing (token already exists), and stores `my-task:2` in resourceState
* I wait long enough
* The controller no longer finds tasks for `my-task:1`
* The controller calls `resource.Delete` (for `my-task:1`)
* The controller deletes the ACL token in Consul for `my-task:1` and clears the AWS Secret
* Next iteration, `reconcile` is called. The controller still has `my-task:2` in resourceState, so Upsert is not called and the token is not recreated. Then, I restart tasks which use the `my-task:2` TaskDefinitionARN, which can't start without their ACL token.

controller/resource.go
pglass: Yeah, I think I'm fine with it as-is.

Is there another way for the controller to determine if something is a mesh-task, so that we don't have to fetch tags 🤔 ? Then we could get away with using `ListTaskDefinitions` and everything would be a littler nicer.

controller/resource.go
```diff
@@ -174,17 +176,57 @@ func (t *Task) Upsert() error {
...
 	return nil
 }
 
 func (t *Task) Delete() error {
+	if !t.isMeshTask() {
+		t.Log.Info("skipping non-mesh task", "id", *t.Task.TaskDefinitionArn)
+		return nil
+	}
+
+	tokenList, _, err := t.ConsulClient.ACL().TokenList(nil)
```
pglass, line 193: Would this list all tokens in Consul? We are potentially doing this multiple times per reconcile if they churn a lot of tasks at once.

Could we move this out of `Delete`, so that we are sure to call TokenList at most once per `reconcile`?

controller/resource_test.go
```diff
@@ -186,6 +186,144 @@ func TestTask_Upsert(t *testing.T) {
...
+		createExistingToken                bool
+		updateExistingSecret               bool
+		task                               ecs.Task
+		expectTokenDeletedAndSecretEmptied bool
+	}{
+		"the token for service doesn't exist in consul and the secret is empty": {
+			createExistingToken:  false,
+			updateExistingSecret: false,
+			task: ecs.Task{
+				TaskArn:           pointerToStr("task"),
```
pglass, line 200: nit: [`aws.String`](https://docs.aws.amazon.com/sdk-for-go/api/aws/#hdr-Value_and_Pointer_Conversion_Utilities) gives you a pointer

controller/controller.go
ishustava: Great question! 

I don't think this will be problematic. When you stop a task, you're effectively changing its desired status from `RUNNING` to `STOPPED`. By default, `ListTasks` will return tasks with desired status of `RUNNING`, and so listing tasks will not return the stopped tasks from step 1.

controller/resource_test.go
ishustava: yeah great call!

controller/resource.go
ishustava: I actually realized that we can't list task definitions by cluster, so I don't think we can use it.

controller/resource.go
pglass: Yeah, we can revisit later!

controller/controller.go
ishustava: good question, I'll try to reproduce, although off the top of my head I'm not sure how it could be possible to start two tasks with different task definition versions. I'll try to test out different scaling scenarios and see if I run into any edge cases. 

Thanks for thinking this through ❤️

pglass: > although off the top of my head I'm not sure how it could be possible to start two tasks with different task definition versions

An ECS Service does a rolling update by default (I think). So if you have multiple tasks per service, you could have two task definition versions running during a deployment.

[Docs](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/update-service.html):

> If a service is using the rolling update (ECS) deployment type, the minimum healthy percent represents a lower limit on the number of tasks in a service that must remain in the RUNNING state during a deployment, as a percentage of the desired number of tasks (rounded up to the nearest integer). ... For example, if your service has a desired number of four tasks and a minimum healthy percent of 50%, the scheduler may stop two existing tasks to free up cluster capacity before starting two new tasks.

The other way is to not use services. Just call `RunTask` and pass it the task definition versions.

ishustava: Great catch!!! I tested it with service updates, and it does delete the token like you described. I've updated the code to skip token deletion if there's a service registered with Consul. I haven't found that there will be a race condition when you actually delete the service in my testing, but if I'm missing something this is still safer since it's better to leave tokens around then to delete and cause service to not be able to restart.

controller/resource.go
```diff
@@ -174,17 +176,66 @@ func (t *Task) Upsert() error {
...
+		return fmt.Errorf("parsing service name: %w", err)
+	}
+
+	services, _, err := t.ConsulClient.Catalog().Service(serviceName, "", nil)
+	if err != nil {
+		return fmt.Errorf("getting service from Consul: %w", err)
+	}
+	if len(services) != 0 {
+		t.Log.Info("skipping because service still exists", "id", *t.Task.TaskDefinitionArn)
+		return nil
```
pglass, line 204: 👍 

Only thought: We return `nil` here, which causes this TaskDefinitionARN to be removed from `c.resourceState`. So maybe there's risk the controller forgets to clean up a token... It would be a narrow window, for example:

* The last task of a task family is stopped
* consul-client begins to shutdown
* `reconcile` runs and does not find the stopped task, so calls Delete
* Delete sees the service in Consul, and skips the token cleanup
* The TaskDefinitionARN is removed from c.resourceState
* consul-client removes the service from Consul, and then leaves the cluster

Maybe worth a comment in the code, but it seems rare. And it wouldn't matter if you later start more tasks for the same family, since they would reuse the token.

So I am good with this!

pglass (approved):
Looks great! Thanks for bearing with my review comments! 😄

controller/resource.go
ishustava: > Only thought: We return nil here, which causes this TaskDefinitionARN to be removed from c.resourceState. So maybe there's risk the controller forgets to clean up a token.

Yeah, this is intentional. I wasn't able to create a race condition that you're describing, but if there is one, then it'd be safer to keep the token. I'll add a comment though!

PR #31, opened 2021-09-23
Example of edge case deleting tokens

lkysow:


controller/mocks/sm_client.go
```diff
@@ -7,7 +7,8 @@ import (
 
 type SMClient struct {
 	secretsmanageriface.SecretsManagerAPI
-	Secret *secretsmanager.GetSecretValueOutput
+	Secret        *secretsmanager.GetSecretValueOutput
+	UpdateSecretF func(input *secretsmanager.UpdateSecretInput) (*secretsmanager.UpdateSecretOutput, error)
```
lkysow, line 11: added to allow mocking out this call

PR #33, opened 2021-10-05
Add health checking subcommand

erichaberkorn:
- Add a health sync subcommand for our defaulted TTL checks. I'll create the associated checks in mesh-init in a follow-up PR. I'm adding this first because adding the TTL checks in mesh-init is a breaking change without this and this allows this feature to be shipped iteratively.

## How I've tested this PR:
- There is pretty good test coverage in subcommand/health-sync/command_test.go

Everything should be covered by the tests

## Checklist:
- [ X ] Tests added
- [ X ] CHANGELOG entry added


subcommand/health-sync/command.go
```diff
@@ -0,0 +1,183 @@
...
+}
+
+func (c *Command) syncChecks(currentStatuses map[string]string, serviceID string, parsedContainerNames []string) map[string]string {
+	taskMeta, err := awsutil.ECSTaskMetadata()
+	if err != nil {
+		c.log.Error(err.Error())
+	}
+	containersToSync, missingContainers := findContainersToSync(parsedContainerNames, taskMeta)
+	for _, name := range missingContainers {
+		c.UI.Error(fmt.Sprintf("container %s not found in task metadata", name))
```
pglass, line 102: nit: Make this a warning instead (`c.UI.Warn`)?

subcommand/health-sync/command.go
```diff
@@ -0,0 +1,183 @@
...
+				continue ContainerNames
+			}
+		}
+		missing = append(missing, container)
+	}
+	return ecsContainers, missing
+}
+
+func ecsHealthToConsulHealth(ecsHealth string) string {
+	if ecsHealth != "HEALTHY" {
```
pglass, line 149: nit: a comment here that lists the ECS health statuses would be good (`UNKNOWN`, `HEALTHY`, or `UNHEALTHY`) so it's clearer why it's a binary choice of either critical or passing for the Consul check.

subcommand/health-sync/command_test.go
```diff
@@ -0,0 +1,261 @@
...
+			require.NoError(t, err)
+			t.Cleanup(func() {
+				_ = server.Stop()
+				os.Unsetenv("CONSUL_HTTP_ADDR")
+			})
+
+			consulClient, err := api.NewClient(&api.Config{Address: server.HTTPAddr})
+			require.NoError(t, err)
+			// We need to set this so that consul connect envoy -bootstrap will talk to the right agent.
+			os.Setenv("CONSUL_HTTP_ADDR", server.HTTPAddr)
```
pglass, line 170: nit: This env var is technically not required for this test, since `consulClient` is passed to the Command and knows the server address. And, since we aren't spawning the consul cli in a subprocess (like mesh-init does)

subcommand/health-sync/command_test.go
```diff
@@ -0,0 +1,261 @@
...
+		if !container.essential {
+			continue
+		}
+
+		err = consulClient.Agent().CheckRegister(&api.AgentCheckRegistration{
+			Name:      makeCheckID(serviceID, container.name),
+			ServiceID: serviceID,
+			AgentServiceCheck: api.AgentServiceCheck{
+				TTL: "10000h",
+			},
```
pglass, line 89: I'd like to include the list of checks in the `ServiceRegister` request, instead of in a separate `CheckRegister` request. So that it exactly matches the way mesh-init creates the checks.

Normally, I probably wouldn't care, and it may not matter here, but there was a recent bug I found in Consul depending on how the checks are registered ([Slack thread](https://hashicorp.slack.com/archives/C045NJD17/p1632349570455500))

subcommand/health-sync/command_test.go
```diff
@@ -0,0 +1,261 @@
...
+			os.Setenv("CONSUL_HTTP_ADDR", server.HTTPAddr)
+
+			setupServiceAndChecks(t, consulClient, containers)
+
+			essentialContainers := []string{}
+			for _, container := range containers {
+				if container.essential {
+					essentialContainers = append(essentialContainers, container.name)
+				}
+			}
```
pglass, line 179: I'm wondering if we actually need this logic around essential containers in this test? 

My read of the code is that `health-sync -container-names A,B,C` would sync status of containers A, B, and, C - regardless of whether they are essential (right?)

(We agreed that it makes sense for Terraform to pass only the essential containers to health-sync, but `health-sync` does not need to know/care about that.)

subcommand/health-sync/command.go
```diff
@@ -0,0 +1,181 @@
...
+	// Don't need to do anything for now. Just catch the SIGTERM so we don't exit.
+	// And, print when we receive the SIGTERM
+	go func() {
+		for sig := range sigs {
+			c.log.Info("signal received", "signal", sig)
+		}
+	}()
+}
+
+func (c *Command) syncChecks(currentStatuses map[string]string, serviceName string, parsedContainerNames []string) map[string]string {
```
pglass, line 95: A comment on this function would be useful, to describe the health sync process and call out assumptions.

pglass (commented):
Looks great overall! Mostly nitpicky comments from me

subcommand/health-sync/command.go
```diff
@@ -0,0 +1,183 @@
...
+			if err != nil {
+				c.UI.Error(fmt.Sprintf("failed to update Consul health status: %s", err.Error()))
+			} else {
+				c.log.Info(fmt.Sprintf("Container %s health check updated in Consul", container.Name))
+				currentStatuses[container.Name] = container.Health.Status
+			}
+		}
+
+		// The rate limit is about 40 per second, so 1 second pulling seems reasonable
+		time.Sleep(1 * time.Second)
```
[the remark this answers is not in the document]
erichaberkorn: Great call out. This will allow us to exit the loop immediately after the SIGTERM handler updates the health checks.

subcommand/health-sync/command_test.go
erichaberkorn: This is testing the case where there is a container that isn't essential. It [doesn't create a check for non-essential containers](https://github.com/hashicorp/consul-ecs/pull/33/files#diff-d2e144a88d61b2f20b2e5ba950a301a0b6d5328c70979b7713d66ffec95a5b6fR69-R71) and that ensures we aren't trying to set [a TTL status for a non-essential container](https://github.com/hashicorp/consul-ecs/pull/33/files#diff-1d2a3112468dcca20d98917f20cd62cf528be35db0af6f7e1274d599c4bb6a12R179) (which would fail). This test code is a bit more complicated than I would optimally want tests to be, but I think it is testing a legitimate case. Does this make sense to you?

pglass: Oh, okay. I think I understand.

1. You're using that `essential` field purely for test setup, to control which containers are prepared with a corresponding TTL check?
2. My read of the code is that the `health-sync` command will attempt to sync status for all container names given, regardless of whether the containers are essential. Is that intentional? (It seems like the right behavior to me)

Assuming "yes" to both of those, then this is good with me.

subcommand/health-sync/command_test.go
```diff
@@ -0,0 +1,255 @@
...
+			continue
+		}
+
+		require.True(t, ok)
+		require.Equal(t, ecsHealthToConsulHealth(container.status), check.Status)
+	}
+}
+
+var cases map[string][]minimalContainerMetadata = map[string][]minimalContainerMetadata{
+	"one essential healthy container": {{
```
pglass, line 112: This was driving me crazy - I think `go fmt` just needs a nudge.

If you stick a newline like this:

```
    "one essential healthy container": {
    {
```

And then run `go fmt`, it will format like the following:

```
    "one essential healthy container": {
        {
            name:      "container1",
            status:    "HEALTHY",
            essential: true,
        },
    },
```

subcommand/health-sync/command_test.go
erichaberkorn: Yes and yes. Thanks for thinking through this with me 😄

subcommand/health-sync/command_test.go
erichaberkorn: Nice find. `go fmt` isn't advanced enough to deal with my attempt to format Go like Clojure 😆

PR #46, opened 2021-11-10
Fix edge cases in the ACL controller when things could fail and never recover

erichaberkorn:
    
Since ACL tokens are used to determine if the reconcile process needs to do anything, there is a bug when ACL tokens are deleted or updated but the Secrets Manager secrets aren't. This is fixed by making the `Upsert` function only insert a secret once and then no longer default it again in the `Delete` function.  This means that after the initial insert, updating and deleting tokens only requires a single operation and can't end up in a partially inserted or deleted state.  The originally insert still takes two operations because it requires inserting the AWS secret and then inserting the ACL token into Consul. Interesting cases:
* The only failure case occurs if the AWS secret is successfully inserted and then the ACL token fails to be deleted. When this happens, a subsequent reconcile will see that the AWS secret has been inserted and then reuse that secret when setting the `AccessorID` and `SecretID` on the ACL token it inserts.
* If a delete fails to delete the ACL token and then then tasks for that service start again, the secret and token already be set correctly and nothing will need to be reconciled for the service to run.
* If an upsert fails to insert the ACL token and then all tasks for the service are stopped, the token will be deleted in the next reconcile.

## How I've tested this PR:
My 🧠 and a new test

👀 and 🧠 

## Checklist:
- [x] Tests added
- [x] CHANGELOG entry added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,8 +21,8 @@ FEATURES
   to support graceful shutdown. [[GH-42](https://github.com/hashicorp/consul-ecs/pull/42)]
 
 BUG FIXES:
-* Fix edge cases in the ACL controller where ACL tokens never get cleaned
-  up. [[GH-45](https://github.com/hashicorp/consul-ecs/pull/45)]
+* Fix bugs in which ACL tokens are not created or deleted in certain cases.
+  [[GH-45](https://github.com/hashicorp/consul-ecs/pull/45)] [[GH-46](https://github.com/hashicorp/consul-ecs/pull/46)]
 
 ## 0.2.0-beta2 (September 30, 2021)
 IMPROVEMENTS
--- a/controller/resource.go
+++ b/controller/resource.go
@@ -12,6 +12,7 @@ import (
 	"github.com/aws/aws-sdk-go/service/secretsmanager/secretsmanageriface"
 	"github.com/hashicorp/consul/api"
 	"github.com/hashicorp/go-hclog"
+	"github.com/hashicorp/go-uuid"
 )
 
 const meshTag = "consul.hashicorp.com/mesh"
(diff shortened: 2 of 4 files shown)


controller/resource.go
```diff
@@ -276,48 +262,74 @@ func (s *ServiceInfo) Delete() error {
 		s.Log.Info("token deleted successfully", "service", s.ServiceName)
 	}
 
+	return nil
+}
+
+// upsertSecret updates the AWS secret for the given service has a Token and
+// AccessorID if it is unset. If the secret is already set, this does nothing.
+func (s *ServiceInfo) upsertSecret() (tokenSecretJSON, error) {
```
erichaberkorn, line 270: @pglass had the awesome idea of defaulting this in github.com/hashicorp/terraform-aws-consul-ecs. Given the current constraints, I think making this change here probably makes more sense. Although, doing the change in the future will require making a breaking change. What do you all think?

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ BREAKING CHANGES
 * mesh-init: The `-envoy-bootstrap-file` option is removed, and replaced with `-envoy-bootstrap-dir`.
   The Envoy bootstrap config file is written to `envoy-bootstrap.json` within that directory.
   [[GH-42](https://github.com/hashicorp/consul-ecs/pull/42)]
+* acl-controller: Service tokens should be prepopulated with `accessor_id` and
```
pglass, line 10:
```suggestion
* acl-controller: AWS Secrets for a service token must be prepopulated with `accessor_id` and
```

Right?

controller/resource.go
```diff
@@ -276,48 +261,43 @@ func (s *ServiceInfo) Delete() error {
...
+		return currSecret, fmt.Errorf("unmarshalling secret value: %w", err)
+	}
+	return currSecret, nil
 }
 
-// updateServiceToken create a token in Consul and updates AWS secret with token's contents.
-func (s *ServiceInfo) updateServiceToken(secretName string) error {
+// updateServiceToken inserts an ACL token into Consul. The AccessorID and
+// SecretID are set based on the AWS secret.
+func (s *ServiceInfo) updateServiceToken(secret tokenSecretJSON) error {
```
pglass, line 287:
```suggestion
func (s *ServiceInfo) createServiceToken(secret tokenSecretJSON) error {
```

Since it is directly creating the token.

pglass (approved):
This is so much nicer 🎉

controller/resource.go
```diff
@@ -225,31 +226,15 @@ func (s *ServiceInfo) Reconcile() error {
...
-		// Read the token with this Accessor ID from Consul.
-		currToken, _, err = s.ConsulClient.ACL().TokenRead(currSecret.AccessorID, nil)
-
-		if err != nil && !isACLNotFoundError(err) {
-			return fmt.Errorf("reading existing token: %w", err)
-		}
+	currToken, _, err := s.ConsulClient.ACL().TokenRead(currSecret.AccessorID, nil)
+
+	if err != nil && !isACLNotFoundError(err) {
+		return fmt.Errorf("reading existing token: %w", err)
```
erichaberkorn, line 237: I'm leaving this around since there is a test for this code and it isn't causing any problems, but it really shouldn't be necessary. During this same reconcile cycle, a token is verified to not exist.

pglass (approved):

lkysow (approved):

controller/resource.go
```diff
@@ -259,14 +243,15 @@ func (s *ServiceInfo) Upsert() error {
...
+	err = s.createServiceToken(currSecret)
 	if err != nil {
 		return fmt.Errorf("updating service token: %w", err)
 	}
 
 	return nil
 }
 
+// Delete removes the token for the given ServiceInfo.
 func (s *ServiceInfo) Delete() error {
```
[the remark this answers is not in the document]
erichaberkorn: I've considered making this change, but am not 100% sure it is an overall positive. I'm imagining a scenario where Consul is in a bad state and fails to process a ACL token deletion. Does it make sense to continue deleting tokens when we know that it [will retry in 10 seconds](https://github.com/hashicorp/consul-ecs/blob/main/controller/controller.go#L11) anyway? Also, the loop here is just for the case where two ACL tokens have been generated by the ACL controller for the same service which I think will only happen if two ACL controllers are running and making tokens for the same service at the same time.

I'm not sure if others agree, but my overall philosophy for the ACL controller is to give up if anything goes wrong because insuring the stability of Consul is more important than making tokens faster.

What do you think?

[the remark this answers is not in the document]
erichaberkorn: At the very least, it feels like we should backlog adding a README for the ACL controller that describes how it works and it's overall philosophy.

PR #53, opened 2021-12-08
File-based config

pglass:
* Support file-based configuration for mesh-init and health-sync

## How I've tested this PR:
Unit tests, and TF acceptance tests in https://github.com/hashicorp/terraform-aws-consul-ecs/pull/77

👀 , but careful review of our config types and JSON Schema fields to make sure all the fields seem right.

## Checklist:
- [x] Tests added
- [x] CHANGELOG entry added


config/types.go
```diff
@@ -0,0 +1,319 @@
...
+
+// SidecarProxyRegistration configures the sidecar proxy registration.
+//
+// NOTE:
+// - The Kind and Port are set by mesh-init, so these fields are not configurable.
+// - The ID, Name, Tags, and Meta are inferred or copied from the service registration by mesh-init.
+// - The bind address will always be localhost in ECS, so the Address and SocketPath are excluded.
+// - The Connect field is excluded. Since the sidecar proxy is being used, it's not a Connect-native
+//   service, and we don't need the recursive proxy config included in the Connect field.
+// - The Namespace field is excluded. mesh-init will use the namespace from the service registration.
```
erichaberkorn, line 100: I love this comment

config/resources/test_extensive_config.json
```diff
@@ -0,0 +1,126 @@
...
+      "tags": ["frontend"],
+      "port": 8080,
+      "address": "127.0.0.1",
+      "socketPath": "/path/to/socket",
+      "taggedAddresses": {
+        "lan": {
+          "address": "192.168.1.5",
+          "port": 8080
+        }
+      },
```
lkysow, line 23: Not sure these make sense since the addresses can't be static, they need to be dynamic at runtime and also if someone set this to for example the wrong IP then the mesh will be broken.

config/resources/test_extensive_config.json
```diff
@@ -0,0 +1,126 @@
...
+          "body": "{\"method\": \"health\"}",
+          "notes": "Health check for the frontend service",
+          "header": {
+            "Content-Type": ["application/json"]
+          },
+          "interval": "30s",
+          "timeout": "10s",
+          "successBeforePassing": 3,
+          "failuresBeforeCritical": 4,
+          "deregisterCriticalServiceAfter": "60m"
```
lkysow, line 48: Not sure if we should offer this since the health check work handles this so this might throw a wrench into things.

config/resources/test_extensive_config.json
```diff
@@ -0,0 +1,126 @@
...
+              "listenerPort": 20001,
+              "path": "/things",
+              "localPathPort": 8080,
+              "protocol": "http2"
+            }
+          ]
+        }
+      }
+    }
+  }
```
lkysow, line 125: should we support `token`? For folks that want to hardcode a token.

lkysow (commented):
Couple more comments on the schema

config/resources/test_extensive_config.json
pglass: Ooh, yes. I think so. 

I'll make that change in a follow up PR, just because this PR is pretty large already. It should be simple though.

config/resources/test_extensive_config.json
pglass: Ah, I removed this for the proxy but not the service. I'm thinking `socketPath` something we should remove here as well?

lkysow: yeah I think so

config/resources/test_extensive_config.json
lkysow: maybe it will need to be under the top-level acls section though because it kind of feeds in to everything

config/gen.go
```diff
@@ -0,0 +1,45 @@
+//go:build ignore
+// +build ignore
+
+// This is used with `go generate` to copy schema.json into schema.go.
```
pglass, line 4: Forgot to `git add` this earlier. It just copies over the schema.json into schema.go, so that we can edit the schema.json directly, use it with other schema processing tools potentially, and etc.

If we change the schema.json, we need to run `go generate ./...` to update schema.go, and then commit the file.

I haven't added a CI check to validate the schema.go is up to date, yet. I left that for a follow up.

config/validate_test.go
```diff
@@ -0,0 +1,212 @@
...
+					},
+					{
+						CheckID: "frontend-ttl",
+						Name:    "frontend-ttl",
+						TTL:     "10m",
+						Status:  "passing",
+					},
+					// TODO: api.AgentServiceCheck has no H2Ping field (in v1.10.1)
+					// 		 Looks like it's been added to the api package on Consul main,
+					// 		 so should come in a future release.
```
pglass, line 158: I noticed this after correcting the schema to disallow additional properties for checks. `H2Ping` doesn't exist in the Consul `api` package, so users can't register an HTTP2 check (but they couldn't before file-based config either).

config/gen.go
```diff
@@ -0,0 +1,45 @@
...
+	"os"
+	"strings"
+)
+
+const (
+	schemaFile = "schema.json"
+	outputFile = "schema.go"
+)
+
+func main() {
```
erichaberkorn, line 21: This is really cool. Much nicer than inline JSON :)

config/schema.json
```diff
@@ -0,0 +1,299 @@
...
+      "additionalProperties": false
+    },
+    "mesh": {
+      "description": "Configure the Consul mesh",
+      "type": "object",
+      "properties": {
+        "bootstrapDir": {
+          "type": "string"
+        },
+        "healthSyncContainers": {
```
erichaberkorn, line 48: What do you think about adding descriptions to the fields that don't come from Consul types? Some examples are `healthSyncContainers` and the fields in `aclTokenSecret`. This can be done in another PR though.

config/schema.json
```diff
@@ -0,0 +1,299 @@
...
+          "items": {
+            "type": "string"
+          },
+          "minItems": 1,
+          "uniqueItems": true
+        },
+        "service": {
+          "type": "object",
+          "properties": {
+            "name": {
```
erichaberkorn, line 59: I believe `name` is required and it needs to have a positive length. This will be caught at runtime, but it would be nice to specify this here.

config/types.go
```diff
@@ -0,0 +1,274 @@
...
+
+// ExposePath are the paths to expose outside of connect. See ExposeConfig.
+type ExposePath struct {
+	ListenerPort  int    `json:"listenerPort,omitempty"`
+	Path          string `json:"path,omitempty"`
+	LocalPathPort int    `json:"localPathPort,omitempty"`
+	Protocol      string `json:"protocol,omitempty"`
+}
+
+func (e *ExposePath) ToConsulType() api.ExposePath {
```
erichaberkorn, line 267: This must have been brutal, but I love the way this file looks

config/types_test.go
```diff
@@ -0,0 +1,190 @@
+package config
+
+import (
+	"testing"
+
+	"github.com/hashicorp/consul/api"
+	"github.com/stretchr/testify/require"
+)
+
+func TestServiceRegistrationToConsulType(t *testing.T) {
```
erichaberkorn, line 10: It would be super helpful to have a test that the types round trip rather than just one way. In the past I've done this with property based test, but it probably isn't practical in Go.

This is a random thought and probably not actionable

config/validate.go
```diff
@@ -0,0 +1,52 @@
...
+	"encoding/json"
+	"fmt"
+	"os"
+
+	"github.com/hashicorp/go-multierror"
+	"github.com/xeipuuv/gojsonschema"
+)
+
+const (
+	ConfigEnvironmentVariable = "CONSUL_ECS_CONFIG_JSON"
```
erichaberkorn, line 13: I believe this can be private

config/validate.go
```diff
@@ -0,0 +1,52 @@
...
+		return nil
+	}
+
+	for _, e := range result.Errors() {
+		err = multierror.Append(err, fmt.Errorf("%s", e.String()))
+	}
+	return err
+}
+
+func Parse(encodedConfig string) (*Config, error) {
```
erichaberkorn, line 34: I think this can be private too. I almost certainly am at fault for this 😆

config/validate_test.go
```diff
@@ -0,0 +1,212 @@
...
+			parsedConfig, err := Parse(rawConfig)
+			require.NoError(t, err)
+			require.Equal(t, c.expectedConfig, parsedConfig)
+		})
+	}
+}
+
+func TestParseErrors(t *testing.T) {
+	rawConfig := OpenFile(t, "resources/test_config_missing_fields.json")
+	// TODO test multiple errors
```
erichaberkorn, line 39: Should we get rid of this?

subcommand/mesh-init/command.go
```diff
@@ -1,89 +1,61 @@
...
+	config, err := config.FromEnv()
+	if err != nil {
+		c.UI.Error(fmt.Sprintf("invalid config: %s", err))
 		return 1
 	}
-	if c.flagEnvoyBootstrapDir == "" {
-		c.UI.Error(fmt.Sprintf("-%s must be set", flagEnvoyBootstrapDir))
+	c.config = config
+
+	if c.config.Mesh.BootstrapDir == "" {
```
erichaberkorn, line 45: What do you think about moving this check to JSON Schema? I like the idea of doing as many validations there as possible.

subcommand/mesh-init/command.go
```diff
@@ -1,89 +1,61 @@
 package meshinit
```
erichaberkorn, line 1: Extracting validations and constructing Consul types makes this much more readable 😄

config/resources/test_extensive_config.json
```diff
@@ -0,0 +1,114 @@
...
+    "healthSyncContainers": [
+      "frontend"
+    ],
+    "service": {
+      "name": "frontend",
+      "tags": [
+        "frontend"
+      ],
+      "port": 8080,
+      "address": "127.0.0.1",
```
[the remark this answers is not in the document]
erichaberkorn: I can't see any reason to specify this either. I vote for removing it

config/types_test.go
erichaberkorn: Yeah, I'm talking about `JSON -> Our Types -> Consul`. In the past, I've found tons of bugs with tests like this. I had no idea 'testing/quick' existed. That's really cool.

config/gen.go
```diff
@@ -0,0 +1,45 @@
...
+	// generated source should have a line that matches the following
+	// regular expression (in Go syntax):
+	//
+	//     ^// Code generated .* DO NOT EDIT\.$
+	//
+	// This line must appear before the first non-comment, non-blank
+	// text in the file.":
+	content := "// Code generated by gen.go. DO NOT EDIT.\n\n" +
+		"package " + os.Getenv("GOPACKAGE") + "\n\n" +
+		"var schema = `" + strings.TrimSpace(string(schema)) + "`\n"
```
lkysow, line 37: could we use `go:embed` to get `schema.json` available as a variable instead?

lkysow (approved):
This looks really good!

Later it would be good to add a checklist to contributing.md which has high-level steps for what files/tests should be updated when we add a new key to the config.

config/resources/test_extensive_config.json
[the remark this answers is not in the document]
lkysow: yep agreed, footgun

config/gen.go
lkysow: see my comment about go:embed

config/gen.go
pglass: > could we use go:embed 

Oh, TIL. Yeah, that's much simpler.

config/gen.go
pglass: Yep, go:embed is perfect for this. Since `go build` handles go:embed, the schema content is committed in only one place (schema.json), no custom script, no `go generate` or any extra build commands, and no CI check needed to ensure the schema.go is up to date. 🎉

pglass:
I'd like to merge this, and handle a few follow ups afterwards:

* Allowing nullable/empty values for optional fields to make it easier to use from Terraform or elsewhere
* Support a `token` field
* Update the ACL controller to use the file-based config

erichaberkorn (approved):
This looks great. Awesome work :shipit:

PR #58, opened 2022-01-25
Generate reference config

pglass:
* Add a script to generate reference config from the JSON schema (`docgen` directory).
* Add descriptions to all the fields in the schema.json

## How I've tested this PR:

Ran the script, and reviewed the markdown.

## Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -60,4 +60,14 @@ dev-docker: build-dev-dockerfile
 		--build-arg 'GIT_DESCRIBE=$(GIT_DESCRIBE)' \
 		-f $(CURDIR)/build-support/docker/Dev.dockerfile $(CURDIR)
 
-.PHONY: build-image ci.dev-docker dev-docker build-dev-dockerfile
+# Generate reference config documentation.
+# Usage:
+#   make reference-configuration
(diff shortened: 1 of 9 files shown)


docgen/main.go
```diff
@@ -0,0 +1,153 @@
+// Generate markdown from the JSON schema for the Consul ECS config.
```
lkysow, line 1: in the k8s project i've been putting these things inside a `hack` folder. What do you think?

docgen/main.go
```diff
@@ -0,0 +1,153 @@
...
+package main
```
lkysow, line 2: Can you put a comment at the top here (or a readme in this dir) that gives a brief description of what this script does and how to run it.

lkysow (approved):
Should we add a Makefile target too?

docgen/main.go
pglass: Oh, fine with me. Maybe like `hack/generate-config-reference`?

lkysow: sounds good!

PR #61, opened 2022-03-01
Admin Partition Support

cthain:
- Implement support for partitions and namespaces (Consul Enterprise)

## How I've tested this PR:
- Augment existing unit tests and add new unit tests for enterprise features.

- Test against Consul: `go test ./...`
- Test against Consul Enterprise: `go test ./... -- -enterprise` 

## Checklist:
- [x] Tests added
- [x] CHANGELOG entry added

### DIFF
--- a/controller/resource_test.go
+++ b/controller/resource_test.go
@@ -37,93 +37,57 @@ import (
 
 func TestServiceStateLister_List(t *testing.T) {
 	enterprise := enterpriseFlag()
-	definitionArn := aws.String("arn:aws:ecs:us-east-1:1234567890:task-definition/service1:1")
 	cluster := "cluster"
 	meshKey := "consul.hashicorp.com/mesh"
 	meshValue := "true"
 	partitionKey := "consul.hashicorp.com/partition"
-	partitionValue := "default"
-	extPartitionValue := "external-partition"
 	namespaceKey := "consul.hashicorp.com/namespace"
 	namespaces := []string{"default", "namespace-1"}
 	meshTag := &ecs.Tag{Key: &meshKey, Value: &meshValue}
-	partitionTag := &ecs.Tag{Key: &partitionKey, Value: &partitionValue}
-	extPartitionTag := &ecs.Tag{Key: &partitionKey, Value: &extPartitionValue}
-	namespace1Tag := &ecs.Tag{Key: &namespaceKey, Value: &namespaces[0]}
-	namespace2Tag := &ecs.Tag{Key: &namespaceKey, Value: &namespaces[1]}
-	task1 := ecs.Task{
-		TaskArn:           aws.String("task1"),
-		TaskDefinitionArn: definitionArn,
-		Tags:              []*ecs.Tag{meshTag},
-	}
-	task1Name := ServiceName{Name: "service1"}
-	task2 := ecs.Task{
-		TaskArn:           aws.String("task2"),
-		TaskDefinitionArn: definitionArn,
-		Tags:              []*ecs.Tag{meshTag},
-	}
-	task2Name := ServiceName{Name: "service1"}
-	nonMeshTask := ecs.Task{
-		TaskArn:           aws.String("nonMeshTask"),
-		TaskDefinitionArn: definitionArn,
-	}
-	task3Name := ServiceName{Name: "service3"}
-	task3 := ecs.Task{
-		TaskArn:           aws.String("task3"),
-		TaskDefinitionArn: definitionArn,
-		Tags:              []*ecs.Tag{extPartitionTag},
-	}
-	aclToken1 := &api.ACLToken{
-		Description: fmt.Sprintf("Token for service1 service\n%s: %s", clusterTag, cluster),
-	}
-	aclTokenListEntry1 := &api.ACLTokenListEntry{
-		Description: fmt.Sprintf("Token for service1 service\n%s: %s", clusterTag, cluster),
-	}
-	aclToken2 := &api.ACLToken{
-		Description: fmt.Sprintf("Token for service1 service\n%s: %s", clusterTag, cluster),
-	}
-	aclTokenListEntry2 := &api.ACLTokenListEntry{
-		Description: fmt.Sprintf("Token for service1 service\n%s: %s", clusterTag, cluster),
-	}
-	aclToken3 := &api.ACLToken{
-		Description: fmt.Sprintf("Token for service3 service\n%s: %s", clusterTag, cluster),
-	}
-	aclTokenListEntry3 := &api.ACLTokenListEntry{
-		Description: fmt.Sprintf("Token for service3 service\n%s: %s", clusterTag, cluster),
-	}
+	nonMeshTag := &ecs.Tag{}
 
-	if enterprise {
-		// add partitions and namespaces for enterprise testing.
-		task1Name.Partition = partitionValue
-		task1Name.Namespace = namespaces[0]
-		task1.Tags = append(task1.Tags, partitionTag, namespace1Tag)
-		task2Name.Partition = partitionValue
-		task2Name.Namespace = namespaces[1]
-		task2.Tags = append(task2.Tags, partitionTag, namespace2Tag)
-		task3Name.Partition = partitionValue
-		task3Name.Namespace = namespaces[0]
-		task3.Tags = append(task3.Tags, meshTag, extPartitionTag, namespace1Tag)
-		aclToken1.Partition = partitionValue
-		aclToken1.Namespace = namespaces[0]
-		aclTokenListEntry1.Partition = partitionValue
-		aclTokenListEntry1.Namespace = namespaces[0]
-		aclToken2.Partition = partitionValue
-		aclToken2.Namespace = namespaces[1]
-		aclTokenListEntry2.Partition = partitionValue
-		aclTokenListEntry2.Namespace = namespaces[1]
-		aclToken3.Partition = partitionValue
-		aclToken3.Namespace = namespaces[0]
-		aclTokenListEntry3.Partition = partitionValue
-		aclTokenListEntry3.Namespace = namespaces[0]
+	makeTask := func(taskName, serviceName, partition, namespace string, tags ...*ecs.Tag) (ecs.Task, ServiceName, *api.ACLToken, *api.ACLTokenListEntry) {
+		if enterprise {
+			tags = append(tags, &ecs.Tag{Key: &partitionKey, Value: &partition})
+			tags = append(tags, &ecs.Tag{Key: &namespaceKey, Value: &namespace})
+		}
+		task := ecs.Task{
+			TaskArn:           aws.String(taskName),
+			TaskDefinitionArn: aws.String(fmt.Sprintf("arn:aws:ecs:us-east-1:1234567890:task-definition/%s:1", serviceName)),
+			Tags:              tags,
+		}
+		name := ServiceName{Name: serviceName}
+		token := &api.ACLToken{
+			Description: fmt.Sprintf("Token for %s service\n%s: %s", serviceName, clusterTag, cluster),
+		}
+		tokenListEntry := &api.ACLTokenListEntry{
+			Description: fmt.Sprintf("Token for %s service\n%s: %s", serviceName, clusterTag, cluster),
+		}
+		if enterprise {
+			name.Partition = partition
+			name.Namespace = namespace
+			token.Partition = partition
+			token.Namespace = namespace
+			tokenListEntry.Partition = partition
+			tokenListEntry.Namespace = namespace
+		}
+		return task, name, token, tokenListEntry
 	}
+	task1, task1Name, aclToken1, aclTokenListEntry1 := makeTask("task1", "service1", "default", "default", meshTag)
+	task2, task2Name, aclToken2, aclTokenListEntry2 := makeTask("task2", "service1", "default", "namespace-1", meshTag)
+	nonMeshTask, _, _, _ := makeTask("nonMeshTask", "service1", "default", "default", nonMeshTag)
+	_, task3Name, aclToken3, aclTokenListEntry3 := makeTask("task3", "service3", "default", "default", nonMeshTag)
+	task4, _, _, _ := makeTask("task4", "service4", "external-partition", "default", meshTag)
 	cases := map[string]struct {
 		paginateResults bool
 		tasks           []ecs.Task
 		expected        map[ServiceName]ServiceState
 		aclTokens       []*api.ACLToken
+		partition       string
 	}{
 		"no overlap between tasks, services and tokens": {
-			tasks:     []ecs.Task{task1, task2, task3},
+			tasks:     []ecs.Task{task1, task2},
+			partition: "default",
 			aclTokens: []*api.ACLToken{aclToken3},
 			expected: map[ServiceName]ServiceState{
 				task1Name: {
@@ -136,6 +100,7 @@ func TestServiceStateLister_List(t *testing.T) {
 		},
 		"all overlap between tasks, services and tokens": {
 			tasks:     []ecs.Task{task1, task2},
+			partition: "default",
 			aclTokens: []*api.ACLToken{aclToken1, aclToken2},
 			expected: map[ServiceName]ServiceState{
 				task1Name: {
@@ -150,6 +115,7 @@ func TestServiceStateLister_List(t *testing.T) {
 		},
 		"with pagination": {
 			tasks:           []ecs.Task{task1, task2},
+			partition:       "default",
 			paginateResults: true,
 			expected: map[ServiceName]ServiceState{
 				task1Name: {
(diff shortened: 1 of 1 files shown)


controller/resource.go
```diff
@@ -375,6 +623,37 @@ func tagValue(tags []*ecs.Tag, key string) string {
...
+	return ""
+}
+
+// PartitionsEnabled indicates if support for partitions and namespaces is enabled.
+func PartitionsEnabled(p string) bool {
+	return p != ""
+}
+
+// uniqueNamespaces returns the set of unique namespaces from the list of Resources.
+func uniqueNamespaces(resources []Resource) []string {
```
pglass, line 648: What do you think about returning the set of namespaces (i.e. `map[string]struct{}`)?

controller/resource.go
```diff
@@ -146,48 +179,209 @@ func (s ServiceStateLister) fetchECSTasks() (map[string]struct{}, error) {
...
+func (s ServiceStateLister) createNamespaces(ns []string) error {
+	// retrieve the list of existing namespaces
+	existingNS, _, err := s.ConsulClient.Namespaces().List(&api.QueryOptions{Partition: s.Partition})
+	if err != nil {
+		return err
+	}
+
+	// create any namespaces that do not already exist
+	create := make([]string, 0, len(ns))
+	for _, n := range ns {
```
pglass, line 316: nit: Related to my other comment, if `uniqueNamespaces` returns a set (i.e. `map[string]struct{}`) and this accepts that set, then can we avoid this double nested loop? It might be slightly awkward and not matter much though.

controller/resource.go
```diff
@@ -146,48 +179,209 @@ func (s ServiceStateLister) fetchECSTasks() (map[string]struct{}, error) {
...
+			Name:      n,
+			ACLs:      &api.NamespaceACLConfig{PolicyDefaults: []api.ACLLink{{Name: xnsPolicyName}}},
+		}, nil)
+		if err != nil {
+			s.Log.Error("creating namespace", n, err)
+			failed = append(failed, n)
+		}
+	}
+	if len(failed) > 0 {
+		return fmt.Errorf("namespace creation failed for %s", strings.Join(failed, ","))
```
pglass, line 338: nit: In other places, we've used [`go-multierror`](https://github.com/hashicorp/go-multierror) for collecting multiple errors.

https://github.com/hashicorp/consul-ecs/blob/0ad5fee74c0223f55238b1b80b6cb40f17ed4878/subcommand/health-sync/command.go#L157-L174

controller/resource.go
```diff
@@ -146,48 +179,209 @@ func (s ServiceStateLister) fetchECSTasks() (map[string]struct{}, error) {
...
+	var partition, namespace string
+	if PartitionsEnabled(s.Partition) {
+		partition = tagValue(t.Tags, partitionTag)
+		namespace = tagValue(t.Tags, namespaceTag)
+	}
+	if serviceName := tagValue(t.Tags, serviceNameTag); serviceName != "" {
+		return NewQName(partition, namespace, serviceName), nil
+	}
+	taskDefArn := *t.TaskDefinitionArn
+	splits := strings.Split(taskDefArn, "/")
```
pglass, line 366: nit: fmt.Sscanf could also make this more readable (but I realize you didn't write this portion)

controller/resource.go
```diff
@@ -196,79 +390,112 @@ type ServiceInfo struct {
...
 	return nil
 }
 
-// Upsert creates a token for the task if one doesn't already exist
+// Upsert creates a service policy and token for the task if one doesn't already exist
 // and updates the secret with the contents of the token.
 func (s *ServiceInfo) Upsert() error {
+	opts := &api.QueryOptions{Partition: s.QName.Partition(), Namespace: s.QName.Namespace()}
+
+	// policy
```
pglass, line 425: stray comment

controller/resource.go
```diff
@@ -196,79 +390,112 @@ type ServiceInfo struct {
...
+		}
+		s.Log.Info("policy deleted successfully", "policy", s.policyName())
 	}
 
 	return nil
 }
 
+// GetNamespace returns the namespace that the service belongs to.
+// It returns the empty string if namespaces are not enabled.
+func (s *ServiceInfo) GetNamespace() string {
```
pglass, line 491: nit: `GetNamespace()` -> `Namespace()` to be consistent (and clarify this isn't fetching the namespace). Or, we could access `s.QName.Namespace()` directly, which I see in some places as well.

controller/qname.go
```diff
@@ -0,0 +1,75 @@
+package controller
+
+import (
+	"fmt"
+	"strings"
+)
+
+// QName is a string type that represents a fully qualified name that includes the
+// partition, namespace and service name.
+type QName struct {
```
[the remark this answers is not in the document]
cthain: Agreed. I will change the name from `QName` to `ServiceName` and make it a data struct instead of a "behaviour" struct.

controller/qname.go
```diff
@@ -0,0 +1,75 @@
...
+	}
+
+	return QName{partition: partition, namespace: namespace, name: name}
+}
+
+// Partition returns the partition from the qualified name.
+// It returns the empty string if the QName does not contain
+// a partition and namespace.
+func (q QName) Partition() string {
+	if q.partition != "" && q.namespace != "" {
```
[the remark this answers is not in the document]
cthain: Yes, I agree with using a data struct for the `ServiceName`. I would like to discuss the UX of the task configuration and how we want to handle the different permutations of the presence (or absence) of the tags partition/namespace tags.

controller/resource.go
```diff
@@ -126,18 +154,23 @@ func (s ServiceStateLister) fetchECSTasks() (map[string]struct{}, error) {
...
 				continue
 			}
 
-			resources[serviceName] = struct{}{}
+			if serviceName.Partition() != s.Partition {
+				s.Log.Info("skipping task in external partition", serviceName.Partition(), "task-arn", *task.TaskArn)
+				continue
+			}
+
+			resources[serviceName.String()] = serviceName
```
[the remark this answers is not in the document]
erichaberkorn: That would be awesome. Just to clarify, we should make our own struct that is similar to `ServiceName` rather than import and depend on it. As far as I know, that part of Consul isn't supposed to be used as a library

controller/qname.go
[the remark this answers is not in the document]
cthain: @erichaberkorn @pglass Please confirm that per our meeting the following is the behaviour we will support for the tags on service tasks:

| Task Tag | ACL Controller Behaviour |
| - | - |
| `consul.hashicorp.com/mesh` | Indicates that the task is part of the service mesh. If this tag is not present the task is ignored by the ACL Controller. |
| `consul.hashicorp.com/service-name` | Name of the service. This tag is optional and the value is used if present; otherwise, the service name is inferred from task ARN. |
| `consul.hashicorp.com/partition` | `Enterprise` Admin partition the service belongs to. This tag is optional and if it is not present the `default` partition is used. If it is present the `consul.hashicorp.com/namespace` tag must also be included in the task definition or the ACL controller will log an error and ignore the task. |
| `consul.hashicorp.com/namespace` | `Enterprise` Namespace the service belongs to within the partition. This tag is optional and if it is not present the `default` namespace is used. If it is present the `consul.hashicorp.com/partition` tag must also be included in the task definition or the ACL controller will log an error and ignore the task. |

controller/resource.go
```diff
@@ -196,79 +390,112 @@ type ServiceInfo struct {
...
+// TokenSecretJSON is the struct that represents JSON of the token secrets
 // stored in Secrets Manager.
-type tokenSecretJSON struct {
+type TokenSecretJSON struct {
 	AccessorID string `json:"accessor_id"`
 	Token      string `json:"token"`
 }
 
 // Reconcile inserts or deletes ACL tokens based on their ServiceState.
 func (s *ServiceInfo) Reconcile() error {
```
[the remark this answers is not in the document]
cthain: Thanks, @pglass and I did this part together.

controller/resource.go
```diff
@@ -126,18 +154,23 @@ func (s ServiceStateLister) fetchECSTasks() (map[string]struct{}, error) {
...
+			serviceName, err := s.serviceNameForTask(task)
 
 			if err != nil {
 				s.Log.Error("couldn't get service name from task", "task-arn", task.TaskArn, "tags", task.Tags, "err", err)
 				continue
 			}
 
-			resources[serviceName] = struct{}{}
+			if serviceName.Partition() != s.Partition {
+				s.Log.Info("skipping task in external partition", serviceName.Partition(), "task-arn", *task.TaskArn)
```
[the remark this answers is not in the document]
cthain: Agreed. Thanks for the pointer! I've updated all logging statements accordingly.

controller/resource.go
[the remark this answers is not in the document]
cthain: Moved to new `ServiceName` struct and now use it for the key in the mappings.

controller/resource.go
```diff
@@ -146,48 +179,209 @@ func (s ServiceStateLister) fetchECSTasks() (map[string]struct{}, error) {
...
+				}
 			}
 		}
 	}
 
-	return aclTokens, nil
+	return aclState, nil
+}
+
+// prepareNamespaces ensures that for every service in the cluster the namespace
```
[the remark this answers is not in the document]
cthain: I manually implemented this suggestion on top of my other changes.

controller/resource.go
```diff
@@ -146,48 +179,209 @@ func (s ServiceStateLister) fetchECSTasks() (map[string]struct{}, error) {
...
+			}
+		}
+
+		tokenList, _, err := s.ConsulClient.ACL().TokenList(opts)
+		if err != nil {
+			return aclState, err
+		}
+
+		for _, token := range tokenList {
+			//if isInCluster(s.Cluster, token.Description) && len(token.Policies) == 1 {
```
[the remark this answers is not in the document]
cthain: Yes, that is correct. Good catch on the extraneous commented out code.

controller/resource.go
```diff
@@ -375,6 +623,37 @@ func tagValue(tags []*ecs.Tag, key string) string {
...
+func IsACLNotFoundError(err error) bool {
 	return err != nil && strings.Contains(err.Error(), "Unexpected response code: 403 (ACL not found)")
 }
+
+// serviceNameFromDescription returns the fully qualified service name from the
+// description field of a token or policy created by the ACL controller.
+func serviceNameFromDescription(d string) string {
+	// description is of the form: "<Policy|Token> for <name> service..."
+	parts := strings.Split(d, " ")
+	if len(parts) > 3 {
```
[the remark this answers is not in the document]
cthain: Excellent suggestion. Done.

controller/resource.go
cthain: I think it's a great idea. I wasn't too sure about using "sets" but then I read the Go style guide in the eng handbook yesterday and saw that this is acceptable. I have actually just removed this func since it was only used in one spot and is vestigial now given the updated implementation.

controller/resource.go
cthain: Thanks for the pointer.. I've updated the code to use `go-multierror`

controller/resource.go
cthain: Yes, that was a really good call. I think that using a set has made this function a bit more palatable. Using the `go-multierror` also helped clean up the code.

controller/resource.go
cthain: good eyes.

controller/resource.go
cthain: Agreed. I have renamed `GetNamespace()` to `Namespace()`. This is needed to allow `ServiceInfo` to implement the `Namespace()` func in the `Resource` interface. Which is now used by the controller to `ReconcileNamespaces()` outside of the `List()` function.

controller/resource.go
cthain: I'm going to leave this as is for now... While I agree that it could make the code more readable, I played around with `fmt.Scanf` a bit and it seems to struggle with a minimal format specifier like `"%s/%s:%s"`.. another option could be a `regexp` match with positional results.

.circleci/config.yml
```diff
@@ -67,12 +77,12 @@ jobs:
                unzip consul_"${CONSUL_VERSION}"_linux_amd64.zip -d /home/circleci/bin &&
                rm consul_"${CONSUL_VERSION}"_linux_amd64.zip
           PACKAGE_NAMES=$(go list ./...)
-          gotestsum --junitfile $TEST_RESULTS/gotestsum-report.xml -- -p 4 $PACKAGE_NAMES
+          gotestsum --junitfile $TEST_RESULTS/gotestsum-report.xml -- -p 4 $PACKAGE_NAMES -- << parameters.ENTERPRISE_FLAG >>
 
       - store_test_results:
```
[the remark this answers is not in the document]
cthain: Yes I saw this in the latest pipeline run. I'll give your suggestion a try.

controller/policy.go
```diff
@@ -0,0 +1,47 @@
+package controller
+
+// Service policy constants
+
+const ossServicePolicyTpl = `service "%s" {
+  policy = "write"
+}
+service "<serviceName>-sidecar-proxy" {
```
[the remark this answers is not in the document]
cthain: yikes! good catch.

.circleci/config.yml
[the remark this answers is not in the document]
cthain: Yes, that did the trick. Thanks!

controller/resource_test.go
```diff
@@ -12,127 +31,173 @@ import (
...
 		},
 	}
 
+	if enterprise {
+		// in the enterprise case the services are qualified by their partition and namespace
+		// and thus become unique entries.. add the missing expected service states.
+		e := cases["no overlap between tasks, services and tokens"].expected
+		e[task2Name] = ServiceState{ConsulECSTasks: true}
+		e = cases["with pagination"].expected
+		e[task2Name] = ServiceState{ConsulECSTasks: true}
```
pglass, line 178: Thinking aloud: Does it make sense to support OSS-specific and Enterprise-specific cases?

So, for example, then we could create separate versions of those cases with the unique oss vs ent setup:

* `"with pagination [enterprise-only]"`
* `"with pagination [oss-only]"`

And then skip oss-only or enterprise-only tests where appropriate?

```go
for name, c := range cases {
    if enterprise && strings.Contains("oss-only", name) {
        continue
    }
    if !enterprise && strings.Contains("enterprise-only", name) {
        continue
    }
}
```

(Or even just have a `c.enterpriseOnly` flag). Thoughts? Not all that important to me, tbh, so can ignore if it doesn't make sense.

controller/resource_test.go
```diff
@@ -252,84 +342,85 @@ func TestReconcile(t *testing.T) {
...
+			var policies []*api.ACLPolicy
 			var tokens []*api.ACLToken
-			for _, token := range aclTokens[c.sutServiceName] {
-				tokens = append(tokens, &api.ACLToken{
-					Description:       token.Description,
-					ServiceIdentities: token.ServiceIdentities,
-				})
+			if state, ok := aclState[c.sutServiceName]; ok {
+				for _, policy := range state.ACLPolicies {
+					fmt.Println(">>> policy", policy.Description)
```
pglass, line 367: nit: Use `t.Logf`

controller/resource_test.go
```diff
@@ -595,3 +750,494 @@ func TestParseServiceNameFromTaskDefinitionARN(t *testing.T) {
...
+			}
+			require.NoError(t, err)
+			require.Equal(t, beforePolicies, afterPolicies)
+			require.Equal(t, beforeTokens, afterTokens)
+		})
+	}
+}
+
+// helper func that initializes a Consul test server and returns a Consul API client.
+func initConsul(t *testing.T) *api.Client {
```
pglass, line 1149: 👍 I thought we had this, but it was something I added on my mesh-gateway branch (but slightly different for mesh-init).

https://github.com/hashicorp/consul-ecs/blob/286e28b2aaa6125c2eb7b970c9fd993f7f4aeaba/subcommand/mesh-init/command_test.go#L568-L584

I like these types of helpers that encapsulate a setup and a cleanup to declutter our test functions that tend to be quite long.

pglass (approved):
Overall, looks good!

It's a big PR, so I haven't reviewed all the test cases in close detail. But generally, this looks great.

controller/resource_test.go
cthain: Generally I think this makes sense, yes. The logic for this particular case is not really about skipping the test entirely, but modifying the behaviour in the enterprise case; specifically, that when ECS tasks with the same name are detected in the cluster belonging to separate namespaces, they become unique services. Which is a case that we want to test here I think.

Admittedly, this test is a bit lacking because I didn't fully re-factor it - I simply made it work in both the OSS and ENT cases.. for example, there is currently no policy validation.. but policy validation is done in the `TestTaskLifecycle` test cases.

controller/resource_test.go
cthain: thanks for catching my hacky debug statements :grimacing:

I've removed them.

PR #62, opened 2022-03-11
Admin partition updates

cthain:
- Fix the order of operations for the ACL controller initialization: It needs to create the admin partition before creating the ACL policies and tokens.

## How I've tested this PR:
- Using the updated Consul ECS dev image w/ HCP Consul and ECS.

:eyes:

### DIFF
--- a/subcommand/acl-controller/command.go
+++ b/subcommand/acl-controller/command.go
@@ -88,23 +88,24 @@ func (c *Command) run() error {
 			CAPem: []byte(caCert),
 		}
 	}
-	if c.flagPartitionsEnabled {
-		cfg.Partition = c.flagPartition
-	}
 
 	consulClient, err := api.NewClient(cfg)
 	if err != nil {
 		return err
 	}
 
-	smClient := secretsmanager.New(clientSession, nil)
-
-	err = c.upsertConsulClientToken(consulClient, smClient)
-	if err != nil {
+	if err = c.createPartition(consulClient); err != nil {
 		return err
 	}
 
-	if err = c.createPartition(consulClient); err != nil {
+	smClient := secretsmanager.New(clientSession, nil)
+
+	if c.flagPartitionsEnabled {
+		cfg.Partition = c.flagPartition
+	}
+
+	err = c.upsertConsulClientToken(consulClient, smClient)
+	if err != nil {
 		return err
 	}
 
--- a/subcommand/acl-controller/command.go
+++ b/subcommand/acl-controller/command.go
@@ -111,7 +111,7 @@ func (c *Command) run() error {
 		ConsulClient:         consulClient,
 		Cluster:              cluster,
 		SecretPrefix:         c.flagSecretNamePrefix,
-		Partition:            cfg.Partition,
+		Partition:            c.flagPartition,
 		Log:                  c.log,
 	}
 	ctrl := controller.Controller{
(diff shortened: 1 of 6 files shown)


subcommand/acl-controller/command.go
```diff
@@ -88,23 +88,24 @@ func (c *Command) run() error {
...
+		return err
+	}
+
 	smClient := secretsmanager.New(clientSession, nil)
 
-	err = c.upsertConsulClientToken(consulClient, smClient)
-	if err != nil {
-		return err
+	if c.flagPartitionsEnabled {
+		cfg.Partition = c.flagPartition
```
erichaberkorn, line 104: Since the Consul client is already initialized, does this do anything? Even if it does, I wonder if this is just something that would either cause a bug elsewhere or cover up a bug based on [the description in the code](https://github.com/hashicorp/consul/blob/main/api/api.go#L343-L345).

erichaberkorn: Grepping shows that we specify partition in a bunch of places on requests, so I wonder if this is something that would just cover up or cause a bug.

```
controller/resource_test.go
1150:		ns, _, err := consulClient.Namespaces().List(&api.QueryOptions{Partition: p.Name})
1164:	opts := &api.QueryOptions{Partition: partition}
1196:		&api.QueryOptions{Partition: partition, Namespace: namespace})

controller/resource.go
201:	opts := &api.QueryOptions{Partition: s.Partition}
289:		&api.QueryOptions{Partition: s.Partition, Namespace: DefaultNamespace},
330:	existingNS, _, err := s.ConsulClient.Namespaces().List(&api.QueryOptions{Partition: s.Partition})
445:	opts := &api.QueryOptions{Partition: s.ServiceName.Partition, Namespace: s.ServiceName.Namespace}
```

cthain: No, it does nothing.. The `Config` is initialized and then a copy is stored in the `Client`: https://github.com/hashicorp/consul/blob/main/api/api.go#L725. That's a good catch, I'll remove it.

subcommand/acl-controller/command.go
```diff
@@ -88,26 +88,23 @@ func (c *Command) run() error {
 			CAPem: []byte(caCert),
 		}
 	}
-	if c.flagPartitionsEnabled {
-		cfg.Partition = c.flagPartition
-	}
```
pglass, line 93: I think this is fine (I saw the previous discussion). Are we confident we've passed the partition everywhere it's needed in requests @cthain?

pglass (approved):

subcommand/acl-controller/command.go
cthain: @pglass Yes I'm quite confident that we've got all the boxes ticked w.r.t. adding the partition (and namespace) to the queries were necessary. The unit tests are fairly comprehensive and they fail in the enterprise case when the queries aren't qualified with partition and ns.

cthain:
@pglass @erichaberkorn I think this PR is ready to merge. I've tested that everything is working using the [example I created for the terraform repo](https://github.com/hashicorp/terraform-aws-consul-ecs/tree/cthain-add-ns-at/examples/admin-partitions). Will you please have another look over it since I've made a fair amount of changes since it was last reviewed? Thanks!

pglass (approved):

erichaberkorn (approved):
Looks good 👍

subcommand/acl-controller/command.go
```diff
@@ -143,6 +140,7 @@ func (c *Command) Help() string {
 // A non-nil error is returned if the operation fails.
 func (c *Command) createPartition(consulClient *api.Client) error {
 	if !c.flagPartitionsEnabled {
+		c.flagPartition = ""
```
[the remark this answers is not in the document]
cthain: Yes I agree that it is out of place here. I will move it into the `run()` function.

PR #67, opened 2022-04-04
Support log level flag and config field

pglass:
Add a `-log-level` flag to all commands

## How I've tested this PR:
- Unit tests

👀 

## Checklist:
- [x] Tests added
- [x] CHANGELOG entry added

### DIFF
--- a/subcommand/acl-controller/command.go
+++ b/subcommand/acl-controller/command.go
@@ -14,6 +14,7 @@ import (
 	"github.com/aws/aws-sdk-go/service/secretsmanager/secretsmanageriface"
 	"github.com/hashicorp/consul-ecs/awsutil"
 	"github.com/hashicorp/consul-ecs/controller"
+	"github.com/hashicorp/consul-ecs/logging"
 	"github.com/hashicorp/consul/api"
 	"github.com/hashicorp/go-hclog"
 	"github.com/mitchellh/cli"
@@ -39,6 +40,8 @@ type Command struct {
 	flagSet *flag.FlagSet
 	once    sync.Once
 	ctx     context.Context
+
+	logging.LogOpts
 }
 
 func (c *Command) init() {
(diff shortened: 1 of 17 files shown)


subcommand/acl-controller/command.go
```diff
@@ -48,16 +51,19 @@ func (c *Command) init() {
 	c.flagSet.StringVar(&c.flagPartition, flagPartition, "", "The Consul partition name that the ACL controller will use for ACL resources. If not provided will default to the `default` partition [Consul Enterprise]")
 	c.flagSet.BoolVar(&c.flagPartitionsEnabled, flagPartitionsEnabled, false, "Enables support for Consul partitions and namespaces [Consul Enterprise]")
 
-	c.log = hclog.New(nil)
+	logging.Merge(c.flagSet, c.LogOpts.Flags())
```
pglass, line 54: Borrowed this approach from Consul: https://github.com/hashicorp/consul/blob/319fe485613bed39784252f275f221d117235b78/command/login/login.go#L64-L68

subcommand/mesh-init/command.go
```diff
@@ -25,15 +27,25 @@ type Command struct {
...
+	logging.Merge(c.flagSet, c.LogOpts.Flags())
 }
 
-func (c *Command) Run(_ []string) int {
+func (c *Command) Run(args []string) int {
 	c.once.Do(c.init)
 
+	if err := c.flagSet.Parse(args); err != nil {
+		return 1
+	}
```
pglass, line 45: We use JSON config instead of flags for `mesh-init` and `health-sync`, but we don't support JSON config for all commands (`acl-controller`, `app-entrypoint`, `envoy-entrypoint`).

Should we:

1. Use a `-log-level` flag for all commands?
2. Use a JSON `logLevel` field for `mesh-init` and `health-sync`, and `-log-level` for other commands?
3. Switch to JSON for all commands, and use a JSON `logLevel` field for all commands? (This would be a breaking change though at this point).

subcommand/app-entrypoint/command_unix.go
```diff
@@ -30,28 +31,29 @@ type Command struct {
...
 
 func (c *Command) Run(args []string) int {
 	c.once.Do(c.init)
 
 	// Flag parsing stops just before the first non-flag argument ("-" is a non-flag argument)
 	// or after the terminator "--"
-	err := c.flagSet.Parse(args)
-	if err != nil {
+	if err := c.flagSet.Parse(args); err != nil {
 		c.UI.Error(fmt.Sprint(err))
```
lkysow, line 51: err.Error() work?

subcommand/mesh-init/command.go
lkysow: I think 2. mesh-init and health-sync are supposed to be user facing which is why we switched to json. But acl controller isn't user-facing and the entrypoints would probably be annoying to use if you had to write json.

lkysow (commented):
Looks good to me so far

lkysow (approved):
🌲 🪚

PR #69, opened 2022-04-13
Support auth method in mesh-init

pglass:
This enables mesh-init to login and obtain a service token from Consul's IAM auth method.

- Support `consulLogin.enabled` and `consulLogin.method` options in the config json (among a couple others). Also, add `consulHTTPAddr` and `consulCACertFile` config fields.
- mesh-init is responsible for logging in to obtain the service token, and putting it in the bootstrap directory
- health-sync assumes the service-token file exists and uses (created by mesh-init)

## How I've tested this PR:
- Manually with Terraform in ECS
- Unit test

👀 

## Checklist:
- [x] Tests added
- [x] CHANGELOG entry added

### DIFF
--- a/config/schema.json
+++ b/config/schema.json
@@ -15,6 +15,22 @@
       "type": "string",
       "minLength": 1
     },
+    "authMethod": {
+      "type": "object",
+      "properties": {
+        "enabled": {
+          "description": "Enable login to Consul's AWS IAM auth method to obtain an ACL token. This requires the auth method to be configured on the Consul server, and it the ECS task role to be trusted by the auth method. After login, the token is written to the file `<bootstrapDir>/service-token`.",
+          "type": "boolean"
+        },
+        "loginFlags": {
+          "description": "Additional CLI flags to pass to the `consul login` command. These are appended to `consul login -type aws -token-sink-file <file> -aws-auto-bearer-token -aws-include-identity`. The `-method` flag must be included to specify the auth method name.",
+          "type": ["array", "null"],
+          "items": {
+            "type": "string"
+          }
+        }
+      }
+    },
     "healthSyncContainers": {
       "description": "The names of containers that will have health check status synced from ECS into Consul. Cannot be specified with `service.checks`.",
       "type": ["array", "null"],
--- a/config/schema.json
+++ b/config/schema.json
@@ -24,18 +24,18 @@
       "type": ["string", "null"]
     },
     "consulLogin": {
-      "type": "object",
+      "type": ["object", "null"],
       "properties": {
         "enabled": {
           "description": "Enable login to Consul's AWS IAM auth method to obtain an ACL token. This requires the auth method to be configured on the Consul server, and the ECS task role must be trusted by the auth method. After login, the token is written to the file `<bootstrapDir>/service-token`.",
-          "type": "boolean"
+          "type": ["boolean", "null"]
         },
         "method": {
           "description": "The name of Consul auth method. This is passed as the `-method` option to the `consul login` command. Defaults to `iam-ecs-service-token`.",
           "type": ["string", "null"]
         },
-        "noIncludeEntity": {
-          "description": "Set this to true to remove the `-aws-include-entity` flag from the `consul login` command. By default, the `-aws-include-entity` flag is passed to the `consul login` command. The `-aws-include-entity` flag should be passed if and only if the Consul AWS IAM auth method is configured with `EnableIAMEntityDetails=true`.",
+        "includeEntity": {
+          "description": "Defaults to true. Set this to false to remove the `-aws-include-entity` flag from the `consul login` command. The `-aws-include-entity` flag should be passed if and only if the Consul AWS IAM auth method is configured with `EnableIAMEntityDetails=true`.",
           "type": ["boolean", "null"]
         },
         "extraLoginFlags": {
--- a/config/schema.json
+++ b/config/schema.json
@@ -23,7 +23,7 @@
       "description": "The file path of the Consul server CA certificate.",
       "type": ["string", "null"]
     },
-    "authMethod": {
+    "consulLogin": {
       "type": "object",
       "properties": {
         "enabled": {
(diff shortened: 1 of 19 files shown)


config/schema.json
```diff
@@ -10,6 +10,22 @@
...
+          "description": "Enable login to Consul's AWS IAM auth method to obtain an ACL token. This requires the auth method to be configured on the Consul server, and it the ECS task role to be trusted by the auth method. After login, the token is written to the file `<bootstrapDir>/service-token`.",
+          "type": "boolean"
+        },
+        "loginFlags": {
+          "description": "Additional CLI flags to pass to the `consul login` command. These are appended to `consul login -type aws -token-sink-file <file> -aws-auto-bearer-token -aws-include-identity`. The `-method` flag must be included to specify the auth method name.",
+          "type": ["array", "null"],
+          "items": {
+            "type": "string"
+          }
+        }
```
pglass, line 26: In Terraform, 

```
authMethod = var.acls ? {
  enabled = var.acls
  loginFlags = var.acls ? concat(
    ["-method", var.service_token_auth_method_name],
    ["-http-addr", var.consul_http_addr],
    // Depends on consul-client to write out this file.
    var.tls ? ["-ca-file", "/consul/consul-ca-cert.pem"] : [],
    var.consul_partition != "" ? ["-partition", var.consul_partition] : [],
  ) : []
} : null
```

pglass: This is what I had done for easy testing and such for now. I think @erichaberkorn and I agreed we should split out the important fields instead of the list of flags, and also rethink some of the fields:

* Add top level fields for the consul server, like http addr and ca cert file (preparing for agentless a bit)
* Add a field for the auth method name (`-method <name>`)
* Add partition as a top-level field (and deprecate the nested partition fields). So then it is reused by various nested things (auth method, service, proxy, and gateway).
* Add namespace as a top-level field (however, the auth method will typically exist in the default namespace, so may need its own namespace field too)

subcommand/mesh-init/command.go
```diff
@@ -127,6 +142,51 @@ func (c *Command) realRun() error {
...
+		loginOpts = append(loginOpts, "-aws-include-entity")
+	}
+	if len(c.config.ConsulLogin.ExtraLoginFlags) > 0 {
+		loginOpts = append(loginOpts, c.config.ConsulLogin.ExtraLoginFlags...)
+	}
+
+	err := backoff.RetryNotify(func() error {
+		// We'll get errors until the consul binary is copied to the volume ("fork/exec: text file busy")
+		c.log.Debug("login", "cmd", fmt.Sprint(loginOpts))
+		cmd := exec.Command("consul", loginOpts...)
```
lkysow, line 169: Can we use the Consul API instead of CLI?

lkysow (commented):
We'll need to implement this workaround like k8s did to fix an issue with auth methods and stale tokens: https://github.com/hashicorp/consul-k8s/pull/887/files

subcommand/mesh-init/command.go
pglass: No.

Well, technically, we could. We could copy over [this function](https://github.com/hashicorp/consul/blob/a1eb7744070da3a3d2a7fc4be34dafdbaddf1d88/internal/iamauth/util.go#L39) to generate the [BearerToken](https://www.consul.io/api-docs/acl#login-to-auth-method) for the /acl/login request to Consul. But I didn't think ahead to expose that in some Consul package we could import (consul/api? consul/sdk?), so the CLI does the login for now.

testutil/iamauthtest/arn.go
```diff
@@ -0,0 +1,97 @@
+package iamauthtest
```
pglass, line 1: I copied the things needed for a local AWS test server from Consul and put them here. It's verbatim code from Consul (aside from combining two files). This is what I've done for now. I think eventually we can extract the corresponding package in Consul, or place it in an external package that we can import.

erichaberkorn (approved):

config/schema.json
```diff
@@ -15,6 +15,38 @@
...
+      "properties": {
+        "enabled": {
+          "description": "Enable login to Consul's AWS IAM auth method to obtain an ACL token. This requires the auth method to be configured on the Consul server, and the ECS task role must be trusted by the auth method. After login, the token is written to the file `<bootstrapDir>/service-token`.",
+          "type": "boolean"
+        },
+        "method": {
+          "description": "The name of Consul auth method. This is passed as the `-method` option to the `consul login` command. Defaults to `iam-ecs-service-token`.",
+          "type": ["string", "null"]
+        },
+        "noIncludeEntity": {
```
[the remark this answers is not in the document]
pglass: Yeah, I agree. I opted for this because it was awkward in Golang to have it default to `true`.

At one point I had something like the following. You need a `*bool` to distinguish if the field is present or not in the JSON, but you need an exported field in the struct for the `json` package.

```go
// ConsulLogin configures login options for the Consul IAM auth method.
type ConsulLogin struct {
    Enabled bool   `json:"enabled"`
    Method  string `json:"method"`
    // Exported field is needed for JSON unmarshalling.
    // But, this defaults to true so you must use c.IncludeEntity().
    RawIncludeEntity *bool    `json:"includeEntity"`
    ExtraLoginFlags  []string `json:"extraLoginFlags"`
}

func (c *ConsulLogin) IncludeEntity() bool {
    if c.RawIncludeEntity == nil {
        // not set - default to true
        return true
    }
    return *c.RawIncludeEntity
}
```

Thoughts?

subcommand/mesh-init/command.go
lkysow: Can be done later but I think it would be nice to use the API for this. Execing the consul binary can cause memory issues (we've seen it on k8s) so it's nice if we don't have to.

I think exposing that function in the api package is a good idea.

lkysow (approved):

config/schema.json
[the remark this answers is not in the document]
pglass: I switched this to `includeEntity` with default false, and used a custom `UnmarshalJSON` to set the default.

PR #71, opened 2022-04-21
Update ACL controller to configure IAM auth method

pglass:
This incrementally adds support for the auth method in the ACL controller.

This adds an `-iam-role-path` flag. If set, the ACL controller does the new behavior of ensuring the IAM auth method is created at startup. If not set, it does the "old" behavior of creating the client token at startup. Later, we'll replace/remove the old behavior.

The ACL controller configures the auth method to trust a wildcard arn, like `arn:aws:iam::123456789:role/path/to/roles/*`. This will permit any role at that path, and in the same account as the controller, to login to the auth method.

## How I've tested this PR:
- Unit tests
- Manually with Terraform in ECS

👀 

## Checklist:
- [x] Tests added
- [x] CHANGELOG entry added

### DIFF
--- a/awsutil/awsutil.go
+++ b/awsutil/awsutil.go
@@ -10,6 +10,7 @@ import (
 	"strings"
 
 	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/arn"
 	"github.com/aws/aws-sdk-go/aws/request"
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/hashicorp/consul-ecs/version"
@@ -46,22 +47,22 @@ func (e ECSTaskMeta) TaskID() string {
 }
 
 func (e ECSTaskMeta) AccountID() (string, error) {
-	split := strings.Split(e.TaskARN, ":")
-	if len(split) < 5 {
-		return "", fmt.Errorf("unable to determine AWS account id from Task metadata")
+	a, err := arn.Parse(e.TaskARN)
+	if err != nil {
+		return "", fmt.Errorf("unable to determine AWS account id from Task ARN: %q", e.TaskARN)
 	}
-	return split[4], nil
+	return a.AccountID, nil
 }
 
 func (e ECSTaskMeta) region() (string, error) {
 	// Task ARN: "arn:aws:ecs:us-east-1:000000000000:task/cluster/00000000000000000000000000000000"
 	// https://docs.aws.amazon.com/general/latest/gr/aws-arns-and-namespaces.html
 	// See also: https://github.com/aws/containers-roadmap/issues/337
-	split := strings.Split(e.TaskARN, ":")
-	if len(split) < 4 {
-		return "", fmt.Errorf("unable to determine AWS region from Task metadata")
+	a, err := arn.Parse(e.TaskARN)
+	if err != nil {
+		return "", fmt.Errorf("unable to determine AWS region from Task ARN: %q", e.TaskARN)
 	}
-	return split[3], nil
+	return a.Region, nil
 }
 
 func ECSTaskMetadata() (ECSTaskMeta, error) {
--- a/awsutil/awsutil.go
+++ b/awsutil/awsutil.go
@@ -45,6 +45,14 @@ func (e ECSTaskMeta) TaskID() string {
 	return split[len(split)-1]
 }
 
+func (e ECSTaskMeta) AccountID() (string, error) {
+	split := strings.Split(e.TaskARN, ":")
+	if len(split) < 5 {
+		return "", fmt.Errorf("unable to determine AWS account id from Task metadata")
+	}
+	return split[4], nil
+}
+
 func (e ECSTaskMeta) region() (string, error) {
 	// Task ARN: "arn:aws:ecs:us-east-1:000000000000:task/cluster/00000000000000000000000000000000"
 	// https://docs.aws.amazon.com/general/latest/gr/aws-arns-and-namespaces.html
--- a/subcommand/acl-controller/command.go
+++ b/subcommand/acl-controller/command.go
@@ -6,6 +6,7 @@ import (
 	"flag"
 	"fmt"
 	"os"
+	"strings"
 	"sync"
 
 	"github.com/aws/aws-sdk-go/aws"
@@ -21,16 +22,23 @@ import (
 )
 
 const (
+	flagIAMRolePath           = "iam-role-path"
 	flagConsulClientSecretARN = "consul-client-secret-arn"
 	flagSecretNamePrefix      = "secret-name-prefix"
 	flagPartition             = "partition"
 	flagPartitionsEnabled     = "partitions-enabled"
 
 	consulCACertEnvVar = "CONSUL_CACERT_PEM"
+
+	// Binding rules don't support a '/' character, so we need compatible IAM role
+	// tag names for the auth method until we can address this in Consul.
+	authMethodServiceNameTag = "consul.hashicorp.com.service-name"
+	authMethodNamespaceTag   = "consul.hashicorp.com.namespace"
 )
 
 type Command struct {
 	UI                        cli.Ui
+	flagIAMRolePath           string
 	flagConsulClientSecretARN string
 	flagSecretNamePrefix      string
 	flagPartition             string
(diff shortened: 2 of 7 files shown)


subcommand/acl-controller/command.go
```diff
@@ -21,16 +22,23 @@ import (
...
 	flagSecretNamePrefix      = "secret-name-prefix"
 	flagPartition             = "partition"
 	flagPartitionsEnabled     = "partitions-enabled"
 
 	consulCACertEnvVar = "CONSUL_CACERT_PEM"
+
+	// Binding rules don't support a '/' character, so we need compatible IAM role
+	// tag names for the auth method until we can address this in Consul.
+	authMethodServiceNameTag = "consul.hashicorp.com.service-name"
+	authMethodNamespaceTag   = "consul.hashicorp.com.namespace"
```
pglass, line 36: Technically, these are new tags. We were not previously adding tags to the task IAM role. So we could use these tag names (`consul.hashicorp.com.service-name`) for the IAM role with no change to the existing tags that we add to the ECS tasks.

subcommand/acl-controller/command.go
```diff
@@ -147,6 +162,101 @@ func (c *Command) Help() string {
...
+		serviceAuthMethod.NamespaceRules = append(serviceAuthMethod.NamespaceRules,
+			&api.ACLAuthMethodNamespaceRule{
+				Selector:      fmt.Sprintf(`entity_tags["%s"] != ""`, authMethodNamespaceTag),
+				BindNamespace: fmt.Sprintf(`${entity_tags.%s}`, authMethodNamespaceTag),
+			},
+		)
+	}
+
+	roleName := "consul-ecs-client-role"
+	policyName := "consul-ecs-client-policy"
```
pglass, line 215: The static role and policy names are okay. If there are multiple controllers in the same datacenter, they can share the same policy. If there are partitions, we'd end up with one role+policy per partition (one per controller), which is fine and necessary.

subcommand/acl-controller/command.go
```diff
@@ -147,6 +162,101 @@ func (c *Command) Help() string {
...
+		return err
+	}
+	path := strings.Trim(c.flagIAMRolePath, "/")
+	if path != "" {
+		path += "/"
+	}
+	boundArnPattern := fmt.Sprintf("arn:aws:iam::%s:role/%s*", account, path)
+
+	clientAuthMethod := &api.ACLAuthMethod{
+		Name:        "iam-ecs-client-token",
```
pglass, line 180: The auth method name should support a name prefix. I'm going to add that in a follow up PR.

The static names work when partitions are enabled, since we assume one controller per partition. If partitions aren't enabled and there are multiple ECS clusters in a single Consul datacenter, then we can either:

1. Have multiple controllers share an instance of the auth method. Each controller updates the `BoundIAMPrincipalARNs` to trust the IAM roles for its account/cluster. There's a race condition there (multiple controllers updating at the same time), but we could mitigate against that somewhat.
2. OR, alternatively, each controller creates its own instance of the auth method, which requires a unique auth method name. Hence, the name prefix.

(1) may not be so bad now that I've written it out, but it is a hair more complex.

subcommand/acl-controller/command.go
erichaberkorn: Based on a quick search, it looks like validations for this are done [here](https://github.com/hashicorp/consul/blob/c786c49282d8b17115fb29aa69d9cf44a1addd0a/agent/consul/acl_endpoint.go#L860). At some point, we might want to standardize all tags on this format, but this makes sense for now.

awsutil/awsutil.go
```diff
@@ -45,6 +45,14 @@ func (e ECSTaskMeta) TaskID() string {
 	return split[len(split)-1]
 }
 
+func (e ECSTaskMeta) AccountID() (string, error) {
```
erichaberkorn, line 48: It might be cleaner to use this https://github.com/aws/aws-sdk-go/blob/main/aws/arn/arn.go#L61-L76

subcommand/acl-controller/command.go
```diff
@@ -185,6 +295,131 @@ var partitionedClientPolicyTpl = `partition "%s" {
...
+
+		if len(role.Policies) == 0 {
+			c.log.Info("updating ACL role with policy", "role", roleName, "policy", policyName)
+			role.Policies = []*api.ACLLink{{Name: policyName}}
+			_, _, err := consulClient.ACL().RoleUpdate(role, c.writeOptions())
+			if err != nil {
+				return fmt.Errorf("updating Consul client ACL role: %s", err)
+			}
+			c.log.Info("update ACL role successfully", "name", roleName)
+		}
```
erichaberkorn, line 356: When would this happen? It looks like policies are included on the role [in a single operation](https://github.com/hashicorp/consul-ecs/pull/71/files#diff-57537ae8c46b9830b6b0ab7933cce6705bf85d9f419f741bcb216178c306f702R361) below.

subcommand/acl-controller/command.go
```diff
@@ -185,6 +295,131 @@ var partitionedClientPolicyTpl = `partition "%s" {
...
+func (c *Command) upsertBindingRule(consulClient *api.Client, bindingRule *api.ACLBindingRule) error {
+	method := bindingRule.AuthMethod
+
+	rules, _, err := consulClient.ACL().BindingRuleList(method, c.queryOptions())
+	if err != nil {
+		return fmt.Errorf("listing ACL binding rules for auth method %s: %w", method, err)
+	}
+	if len(rules) > 0 {
+		// For now, we just expect at least one binding rule to exist.
+		// TODO: Can we create the rule with a client-generated ID?
```
erichaberkorn, line 407: Based on glancing through the code, it looks like we can [set `ID` on the `ACLBindingRule`](https://github.com/hashicorp/consul/blob/main/api/acl.go#L222) and then call `BindingRuleRead` with that `ID` rather than using `BindingRuleList`. I think either approach is probably fine though.

subcommand/acl-controller/command_test.go
```diff
@@ -3,22 +3,307 @@ package aclcontroller
...
+				require.NoError(t, err)
+				for _, part := range partitions {
+					if part.Name != "default" {
+						_, err := consulClient.Partitions().Delete(ctx, part.Name, nil)
+						require.NoError(t, err)
+					}
+				}
+				// Apparently, takes moment for the partition to actually be deleted.
+				retry.RunWith(&retry.Timer{Timeout: 2 * time.Second, Wait: 200 * time.Millisecond}, t, func(r *retry.R) {
+					partitions, _, err = consulClient.Partitions().List(ctx, nil)
```
erichaberkorn, line 175: I believe the first sign that a partition is being deleted is `DeletedAt` being set on partitions.

subcommand/acl-controller/command.go
pglass: Yeah, off the top of my head: The binding rules use both [bexpr](https://github.com/hashicorp/go-bexpr) (in the Selector) and [HIL](https://github.com/hashicorp/hil) templates (in the BindName). bexpr supports JSON pointer syntax so I think any tag should will work there, but it's the HIL templates that support only `[a-zA-Z0-9-.]` characters in variable names.

subcommand/acl-controller/command.go
pglass: Yeah, this a rare what-if scenario. It would happen if the policy is deleted somehow (manually?), then the link to the policy is removed from the role. The way I wrote the unit test made it easy to test for this, so I just handled it.

subcommand/acl-controller/command.go
pglass: Yeah, I think this is a matter of which specific unique ID to use. I think `<account>-<region>-<cluster>` would work to uniqueify these resources (or a hash of that), and that would aslo avoid requiring an extra `-auth-method-prefix` flag.

I'll do that in a follow up though.

subcommand/acl-controller/command_test.go
pglass: Yeah. Would that change the need for a retry loop like this though? I don't think so.

erichaberkorn: Not that it's worth doing, but I think we could eliminate the retry loop by ensuring that the length of `partitions` is one after removing partitions with `DeletedAt` set. That makes the code less understandable so I don't think it's worth it.

pglass: Yeah, I think in that case we'd also need to change the acl-controller code to look for `DeletedAt` as well, and even then I'm not sure if, while the partition is "present" but has `DeletedAt` set, that we could create a partition with the same name?

I think I agree - probably not worth it, but super helpful to think through!

PR #72, opened 2022-04-21
acl-controller: Fix creation of namespaces in a partition

pglass:
When running the controller in a non-default partition, it fails to create namespaces because the client is not passing the partition as a query option.

## How I've tested this PR:
- Unit test

👀 

## Checklist:
- [x] Tests added
- [ ] CHANGELOG entry added

### DIFF
--- a/controller/resource_test.go
+++ b/controller/resource_test.go
@@ -768,6 +768,19 @@ func TestReconcileNamespaces(t *testing.T) {
 				"resource4": {ServiceName: ServiceName{Name: "service-1", Partition: "default", Namespace: "namespace-2"}},
 			},
 		},
+		"with resources in non-default partition": {
+			partition: "part-1",
+			expNS: map[string][]string{
+				"default": {"default"},
+				"part-1":  {"default", "namespace-1", "namespace-2"},
+			},
+			resources: map[string]*ServiceInfo{
+				"resource1": {ServiceName: ServiceName{Name: "service-1", Partition: "part-1", Namespace: "default"}},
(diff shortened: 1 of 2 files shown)


controller/resource_test.go
```diff
@@ -147,6 +147,7 @@ func TestServiceStateLister_List(t *testing.T) {
 	}
 
 	for name, c := range cases {
+		c := c
```
erichaberkorn, line 150: Is this required?

erichaberkorn (approved):

controller/resource_test.go
pglass: Yes, though it's because I recently added `t.Parallel()` to tests in this file (which reduced the test time for this file from ~60s to ~10s on my laptop).

The reason is this gotcha with how Golang closures capture variables (e.g. https://blog.cloudflare.com/a-go-gotcha-when-closures-and-goroutines-collide/). In some cases these tests modify stuff on `c`, and that's not okay when these run in parallel since they actually share a reference to `c` when they run.

erichaberkorn: Makes sense. What do you think about adding a comment at least the first time?

pglass: Yep. There _could_ be other issues with enabling those tests to run in parallel. I'm monitoring it, and if it causes more instability I can revert back to no parallelism there.

lkysow: I'd vote for no parallelism and just eat the 60s because guaranteed someone will miss this.

lkysow (approved):

PR #79, opened 2022-04-27
Mitigate against token replication lag

pglass:
Copy the [consul-k8s approach](https://github.com/hashicorp/consul-k8s/pull/887) for dealing with inherent token replication lag.

## How I've tested this PR:
- Unit test

👀 

## Checklist:
- [x] Tests added
- [x] ~CHANGELOG entry added~ n/a, part of unreleased auth method changes which are already logged

### DIFF
--- a/subcommand/mesh-init/command.go
+++ b/subcommand/mesh-init/command.go
@@ -19,6 +19,8 @@ import (
 
 const (
 	envoyBoostrapConfigFilename = "envoy-bootstrap.json"
+	raftReplicationTimeout      = 2 * time.Second
+	tokenReadPollingInterval    = 100 * time.Millisecond
 )
 
 type Command struct {
@@ -70,6 +72,14 @@ func (c *Command) realRun() error {
 	consulClient, err := api.NewClient(cfg)
 	if err != nil {
 		return fmt.Errorf("constructing consul client: %s", err)
+	}
+
+	if c.config.ConsulLogin.Enabled {
+		// The just-created token is not immediately replicated to Consul server followers.
+		// Mitigate against this by waiting for the token in stale consistency mode.
+		if err := c.waitForTokenReplication(consulClient); err != nil {
+			return err
+		}
 	}
 
 	serviceRegistration, err := c.constructServiceRegistration(taskMeta)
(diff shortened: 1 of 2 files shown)


subcommand/mesh-init/command_test.go
```diff
@@ -573,6 +575,73 @@ func TestConstructLoginCmd(t *testing.T) {
...
+			cmd := &Command{log: hclog.NewNullLogger()}
+			err = cmd.waitForTokenReplication(tokenClient)
+			if c.expError {
+				require.Error(t, err)
+			} else {
+				require.NoError(t, err)
+
+				// Token should exist.
+				_, _, err = tokenClient.ACL().TokenReadSelf(nil)
+				require.NoError(t, err)
```
cthain, line 638: is there any point in also doing the sanity check against the token IDs here?

cthain (approved):
Looks good Paul. I left a minor suggestion.. I'll leave it to you to decide whether or not to implement it.

subcommand/mesh-init/command.go
```diff
@@ -171,6 +181,69 @@ func (c *Command) loginToAuthMethod(tokenFile string, taskMeta awsutil.ECSTaskMe
...
+	// next call to reach different servers and those servers to have different states from each
+	// other, but this is unlikely since clients use sticky connections.
+
+	// Mesh-init talks to the local Consul client agent (for now). We need this to hit the Consul
+	// server(s) directly.
+	newCfg := api.DefaultConfig()
+	newCfg.Address = c.config.ConsulHTTPAddr
+	newCfg.TLSConfig.CAFile = c.config.ConsulCACertFile
+	newCfg.TokenFile = cfg.TokenFile
+	newCfg.Token = cfg.Token
```
pglass, line 221: I don't love the way I "clone" this config. I could change the function to receive a value/copy, e.g. `waitForTokenReplication(cfg api.Config)`, so that the function receives a copy, but there are some pointer fields on the `api.Config` type...

Maybe a common function to construct a new config will help.

pglass: I updated this. I realized it just needed the TokenFile and not the Token as well.

PR #83, opened 2022-05-10
merge mesh gateway PoC

cthain:
- Merge Paul's mesh gateway PoC with latest main
- Updated the unit tests to get them all working

## How I've tested this PR:
- Unit tests

- :eyes: 

## Checklist:
- [x] Tests added
- [ ] CHANGELOG entry added


config/schema.json
```diff
@@ -344,8 +344,89 @@
...
-    }
+    },
+    "gateway": {
+      "description": "Configuration for the gateway proxy registration.",
+      "type": "object",
+      "properties": {
+        "kind": {
+          "description": "The gateway kind.",
+          "type": "string",
+          "enum": ["mesh-gateway", "terminating-gateway", "ingress-gateway"]
```
erichaberkorn, line 355: Since we are working on mesh gateways first, what do you think about removing terminating and ingress gateway for now?

config/types.go
```diff
@@ -7,15 +7,26 @@ import (
...
+	// The consul-ecs binary does not write this file, but health-sync will attempt to do a `consul logout` for this token.
+	ClientTokenFilename = "client-token"
 
-// DefaultAuthMethodName is the default name of the Consul IAM auth method used for `consul login`.
-const DefaultAuthMethodName = "iam-ecs-service-token"
+	// DefaultAuthMethodName is the default name of the Consul IAM auth method used for `consul login`.
+	DefaultAuthMethodName = "iam-ecs-service-token"
+
+	// DefaultGatewayPort (8443) is the default gateway registration port used by 'consul connect envoy -register'.
+	DefaultGatewayPort = 8443
```
erichaberkorn, line 22: Given Consul ECS's architecture, I wonder if there will ever be a good reason to override any of these ports 🤔

cthain: :thinking: I'm not sure that the question is if there is a **_good_** reason to do it.. but at some point someone probably **_will_** want to override it :wink: I think my preference would be to leave it defaulted for now to keep it simple.. we could make it configurable later. Thoughts/Objections?

lkysow: k8s defaults the pod bind port to 8443 but the external port to 443 iirc

subcommand/mesh-init/command.go
```diff
@@ -385,3 +408,45 @@ func (c *Command) constructProxyRegistration(serviceRegistration *api.AgentServi
...
+	gwRegistration.Checks = []*api.AgentServiceCheck{
+		{
+			Name:                           fmt.Sprintf("%s listener", gwRegistration.Kind),
+			TCP:                            net.JoinHostPort(healthCheckAddr.Address, fmt.Sprint(healthCheckAddr.Port)),
+			Interval:                       "10s",
+			DeregisterCriticalServiceAfter: "10m",
+		},
+	}
+
+	c.log.Debug(fmt.Sprintf("%+v", gwRegistration))
```
lkysow, line 449: 👁️

subcommand/mesh-init/command.go
```diff
@@ -385,3 +408,45 @@ func (c *Command) constructProxyRegistration(serviceRegistration *api.AgentServi
...
+	gwRegistration := c.config.Gateway.ToConsulType()
+	gwRegistration.ID = serviceID
+	gwRegistration.Name = serviceName
+	gwRegistration.Meta = mergeMeta(map[string]string{
+		"task-id":  taskID,
+		"task-arn": taskMeta.TaskARN,
+		"source":   "consul-ecs",
+	}, c.config.Gateway.Meta)
+
+	// Health check localhost (default) or the LAN address if specified.
```
lkysow, line 430: might need to use the lan address because not sure envoy binds to 127.0.0.1?

subcommand/mesh-init/command.go
cthain: good spot

PR #84, opened 2022-05-13
Add initial CRT configuration

pglass:

This adds some initial CRT configuration to get "dev" builds working on main. I'll iterate on this some more in follow up changes.

- Add CRT release workflow in build.yml + other CRT config
- Add a top-level Dockerfile with multi-stage builds
- New makefile targets

## How I've tested this PR:

The build.yml workflow has run in Github checks pass. Note that the crt common workflows are only to configured to run for `main` (in .release/ci.yml. Later we'd include release branches as well), so only the build.yml has been exercised and not the CRT common workflows.

Basically, that means the binary/image built from this PR is not yet pushed anywhere yet (other than the github actions artifacts). When merged to main, the built dev artifacts will be pushed Artifactory (and dockerhub under `hashicorppreview`). Staging/production builds must be manually triggered.

👀 

## Checklist:
- [x] Tests added
- [ ] ~CHANGELOG entry added~


.github/workflows/build.yml
```diff
@@ -0,0 +1,294 @@
+name: build
+
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - main
+      - pglass/crt  # TODO: for testing
```
pglass, line 8: For now, I've limited this to just these branches as I onboard. Eventually, I think we should allow this build.yml to run for all branches. Then every pull request builds artifacts in the same way as our release pipeline (without actually publishing them anywhere. That's controlled by the list of release branches in `.release/ci.hcl`)

.release/ci.hcl
```diff
@@ -0,0 +1,237 @@
...
+project "consul-ecs" {
+  // the team key is not used by CRT currently
+  team = "consul-ecs"
+  slack {
+    notification_channel = "C01J8QV0EF8" # team-consul-ecs
+  }
+  github {
+    organization = "hashicorp"
+    repository = "consul-ecs"
+    release_branches = ["main"]
```
pglass, line 12: We'll update this to also include active release branches (i.e. `release/0.5.x`) 

We should start using release branches so that it's easy to do patch releases, and I'll define that process in our release docs soon.

.release/ci.hcl
```diff
@@ -0,0 +1,237 @@
...
+    workflow = "fossa-scan"
+  }
+}
+
+## These are promotion and post-publish events
+## they should be added to the end of the file after the verify event stanza.
+
+event "trigger-staging" {
+// This event is dispatched by the bob trigger-promotion command
+// and is required - do not delete.
```
pglass, line 165: Calling this out: By default, the only automated "promotion" is `promote-dev-docker`.

The staging and production release must be triggered manually.

.release/release-metadata.hcl
```diff
@@ -3,3 +3,4 @@ url_docker_registry_ecr = "https://gallery.ecr.aws/hashicorp/consul-ecs"
 url_license           = "https://github.com/hashicorp/consul-ecs/blob/main/LICENSE"
 url_project_website = "https://www.consul.io/docs/ecs"
 url_source_repository = "https://github.com/hashicorp/consul-ecs"
+url_release_notes = "https://www.consul.io/docs/release-notes"
```
pglass, line 6: I wasn't able to link directly to the consul-ecs section of the release notes, fwiw

Dockerfile
```diff
@@ -0,0 +1,81 @@
...
+ENV PATH="/bin/consul-inject:${PATH}"
+
+VOLUME [ "/consul" ]
+
+# Set up certificates, base tools, and software.
+RUN apk add --no-cache ca-certificates curl gnupg libcap openssl su-exec iputils iptables libc6-compat
+
+USER $BIN_NAME
+ENTRYPOINT ["/bin/consul-ecs"]
+COPY dist/$TARGETOS/$TARGETARCH/$BIN_NAME /bin/
```
pglass, line 65: For now, this copies in the binary built in CI to get something working. Locally, the `make dev` or `make docker/dev` will place the binary here as well.

Currently, we build our release image by downloading the consul-ecs binary from releases.hashicorp.com, which is how we guarantee our released image uses the released binary. I need to do the same here, or figure out if that's no longer need because maybe the CRT promotion process ensures the same thing. I'll do that in a follow up change.

Dockerfile
```diff
@@ -0,0 +1,81 @@
...
+# ===================================
+#
+#   Release images.
+#
+# ===================================
+
+
+# default release image
+# -----------------------------------
+FROM docker.mirror.hashicorp.services/alpine:latest AS release-default
```
pglass, line 16: CRT wants the Dockerfile at the root of the repo, and their examples encourage this style of using multi-stage builds.

Basically, this lets us stick config for multiple images in one Dockerfile. For now, I just have one image ("release-default") to get us started, so all the CRT workflows can run, but this will likely change as I finish things up.

Dockerfile
pglass: Okay, confirmed in Slack that we no longer need to download from releases.hashicorp.com as part of the Docker build to get a matching binary!

The CI-built binary is copied into the image, and then both binary+image are copied through to staging and production for release

erichaberkorn (approved):

PR #88, opened 2022-05-19
Enable race detector during CI tests

pglass:
This enables the race detector and fixes the test issues that trigger it.

## How I've tested this PR:
* CI checks
* Running tests a bunch of times locally with `-race`:

```
gotestsum --format=short-verbose --jsonfile ./go-test-race.log --junitfile ./gotestsum-report.xml -- -race ./... -count=10 -tags=enterprise -- -enterprise
```

👀

## Checklist:
- [x] Tests added
- [ ] CHANGELOG entry added

### DIFF
--- a/controller/controller_test.go
+++ b/controller/controller_test.go
@@ -2,6 +2,7 @@ package controller
 
 import (
 	"context"
+	"sync"
 	"testing"
 	"time"
 
@@ -9,6 +10,9 @@ import (
 	"github.com/hashicorp/go-hclog"
 	"github.com/stretchr/testify/require"
 )
+
+// global mutex for these tests to pass the race detector
+var mutex sync.Mutex
 
 func TestRun(t *testing.T) {
 	t.Parallel()
@@ -36,6 +40,8 @@ func TestRun(t *testing.T) {
 	go ctrl.Run(ctx)
 
 	retry.Run(t, func(r *retry.R) {
+		mutex.Lock()
+		defer mutex.Unlock()
 		require.True(r, lister.nsReconciled)
 		for _, resource := range lister.resources {
 			require.True(r, resource.reconciled)
@@ -54,6 +60,9 @@ type testResource struct {
 }
 
 func (t *testResourceLister) List() ([]Resource, error) {
+	mutex.Lock()
+	defer mutex.Unlock()
+
 	var resources []Resource
 	for _, resource := range t.resources {
 		resources = append(resources, resource)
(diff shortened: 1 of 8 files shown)


controller/controller_test.go
```diff
@@ -36,6 +40,8 @@ func TestRun(t *testing.T) {
 	go ctrl.Run(ctx)
 
 	retry.Run(t, func(r *retry.R) {
+		mutex.Lock()
+		defer mutex.Unlock()
 		require.True(r, lister.nsReconciled)
```
pglass, line 45: The issue here is the `go ctrl.Run()` is concurrently writing the fields that are read here. I opted for just a single global mutex for simplicity here.

controller/resource_test.go
```diff
@@ -243,7 +245,6 @@ func TestTaskStateListerList(t *testing.T) {
 			resources, err := lister.List()
 			require.NoError(t, err)
 
-			sortTaskStates(c.expResources)
```
pglass, line 246: This one uses `t.Parallel()`. So multiple test case were concurrently sorting `c.expResources`, which shares some data across test cases. I move the sort prior to the `t.Run`, so that the sorts don't occur in parallel since the subtest functions are not started until later (I think. This maybe could still be a race, but the race detector seems happy)

subcommand/envoy-entrypoint/command_unix_test.go
```diff
@@ -111,7 +106,7 @@ func TestRun(t *testing.T) {
 				signal.Stop(sigs)
 			})
 
-			ecsMetaRequestCount := 0
```
pglass, line 114: This I just made atomic. You need to use `atomic.Load*` and `atomic.Add*` functions though.

subcommand/app-entrypoint/command_unix_test.go
```diff
@@ -149,16 +142,20 @@ func TestRun(t *testing.T) {
...
-				args = append(args, "/bin/sh", "-c", c.targetCommand)
+				args = append(args, "/bin/sh", "-c", c.fakeApp.Command)
 				exitCodeChan <- cmd.Run(args)
 			}()
 
 			t.Logf("Wait for fake app process to start")
 			retry.RunWith(&retry.Timer{Timeout: 1 * time.Second, Wait: 100 * time.Millisecond}, t, func(r *retry.R) {
-				require.NotNil(r, cmd.appCmd)
-				require.NotNil(r, cmd.appCmd.Process)
-				require.Greater(r, cmd.appCmd.Process.Pid, 0)
```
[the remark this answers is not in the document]
cthain: Ya, I think this is fine. I can't think of an easier way to do this.

testutil/fake-command.go
```diff
@@ -0,0 +1,54 @@
+package testutil
+
+import (
+	"fmt"
+	"path/filepath"
+	"testing"
+)
+
+// fakeCommand is a command/script to be run by tests for "entrypoint" commands. Each command touches
```
cthain, line 9: `FakeCommand` should be capitalized here because it's exported.

cthain (approved):
Just had a minor comment on a missing capital letter in a godoc comment. Good stuff Paul.

PR #90, opened 2022-05-20
Migrate over to Github Actions

pglass:
This moves the dev docker image upload over to Github actions and removes the CircleCI config.

For this PR, the CircleCI config is trimmed down to a disabled workflow since actually removing the CircleCI config file causes a failed PR check. After this PR is merged, we can remove the CircleCI config from the repo (webhook, etc) to stop triggering CircleCI workflows, and then remove the `.circleci/` directory.

## How I've tested this PR:
CI checks, and tested open a PR from a fork

👀 

## Checklist:
- [x] Tests added
- [ ] CHANGELOG entry added


.github/workflows/build.yml
```diff
@@ -246,3 +245,50 @@ jobs:
           dev_tags: |
             hashicorppreview/${{ env.repo }}:${{ env.version }}
             hashicorppreview/${{ env.repo }}:${{ env.version }}-${{ github.sha }}
+
+  upload-dev-docker:
```
samsalisbury, line 249: I'm probably missing some context here, but just curious about the use case for pushing the image directly from this workflow, vs relying on the `dev_tags` specified above? If dev_tags aren't satisfying your use case I wonder if we could make some changes to CRT to better support this project?

.github/workflows/build.yml
```diff
@@ -246,3 +245,50 @@ jobs:
...
+        run: |
+          TAG=hashicorpdev/${{ env.repo }}:${{ env.git-short-sha }}
+          echo "==> Load docker image from tar archive"
+          docker load -i "${{ env.repo }}_${{ env.target }}_linux_${{ matrix.arch }}_${{ env.version }}_${{ github.sha }}.docker.dev.tar"
+
+          echo "==> Tag docker image $TAG"
+          docker tag "hashicorppreview/${{ env.repo }}:${{ env.version }}-${{ github.sha }}" "$TAG"
+
+          echo "==> Docker login"
+          echo ${{ secrets.DOCKER_PASS }} | docker login -u=${{ secrets.DOCKER_USER }} --password-stdin
```
samsalisbury, line 280: In general we're trying to rely less heavily on GitHub secrets and instead entrust CRT common workflows running on our private infra to access needed secrets from Vault to perform work like this. I'm curious if we can better support your use case so we can avoid the need for this?

samsalisbury (approved):
Not blocking this as it all looks reasonable, just some Qs in case we can support this use case better from CRT.

.github/workflows/build.yml
pglass: We have acceptance tests we run pre-merge to test PRs (so for non-release branches, basically). Those tests spin up resources in AWS, so we need a public image that AWS can pull in order to run those tests.

`dev_tags` is great, but the CRT workflows only run for release branches, and they publish "stable" artifacts. Here, I publish an "unstable" image per-commit for all branches. This works well for us - we can test with any commit of any branch - and is quick since it doesn't wait for the sign/notarize steps.

I asked in `#team-rel-eng` about doing this with CRT. We could update the `.release/ci.hcl` in each PR to temporarily include that PR branch in the list of release branches, so that the CRT workflow can run. But, this requires manual steps on every PR, and puts unstable artifacts into the "stable" bucket, which is not ideal.

cthain (approved):
I had a small question but this looks good to me.

.circleci/config.yml
```diff
@@ -5,172 +5,13 @@ executors:
       - image: docker.mirror.hashicorp.services/cimg/go:1.17
 
 jobs:
-  go-fmt-and-vet:
+  disabled_job:
```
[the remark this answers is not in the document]
pglass: We don't need this. I was being too cautious about not disabling CircleCI for all branches before merging this PR. So I had this stub workflow to avoid a failing check from CircleCI on this branch.

I'll remove this and disable the CircleCI webhooks prior to merging this.

PR #93, opened 2022-05-24
Fix test flakiness with TaskMetaServer

pglass:
There's some test flakiness where it appears the fake TaskMetaServer is not ready for requests. See [this job](https://github.com/hashicorp/consul-ecs/runs/6576890687?check_suite_focus=true) and [this job](https://github.com/hashicorp/consul-ecs/runs/6576771628?check_suite_focus=true).

This adds a wait loop to check that the TaskMetaServer server has started before proceeding with tests.

## How I've tested this PR:
CI checks. Retrying the unit tests jobs several times.

👀 

## Checklist:
- [x] Tests added
- [ ] CHANGELOG entry added

### DIFF
--- a/testutil/aws.go
+++ b/testutil/aws.go
@@ -1,15 +1,18 @@
 package testutil
 
 import (
+	"io"
 	"net/http"
 	"net/http/httptest"
 	"os"
 	"testing"
+	"time"
 
 	"github.com/hashicorp/consul-ecs/awsutil"
 	"github.com/hashicorp/consul-ecs/config"
 	"github.com/hashicorp/consul-ecs/testutil/iamauthtest"
 	"github.com/hashicorp/consul/api"
+	"github.com/hashicorp/consul/sdk/testutil/retry"
 	"github.com/stretchr/testify/require"
 )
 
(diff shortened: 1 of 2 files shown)


testutil/aws.go
```diff
@@ -36,12 +47,29 @@ func TaskMetaHandlerFn(t *testing.T, respFn func() string) http.Handler {
 // Because of the environment variable, this is unsafe for running tests in parallel.
 func TaskMetaServer(t *testing.T, handler http.Handler) {
 	ecsMetadataServer := httptest.NewServer(handler)
+
+	// Help detect invalid concurrent servers since this relies on an environment variable.
+	require.Empty(t, os.Getenv(awsutil.ECSMetadataURIEnvVar),
+		"%s already set. TaskMetaServer cannot be used concurrently.", awsutil.ECSMetadataURIEnvVar,
+	)
```
pglass, line 54: This didn't catch any concurrent servers being started. I thought we could keep it anyway as a safeguard / warning?

cthain: Yes, I think this is a really good check to leave in. It will catch any future attempts to use this concurrently.

PR #95, opened 2022-05-25
Disable cgo during builds

pglass:
Disable CGO during builds. This enables us to remove libc6-compat from the image, which is important for our features/workarounds where we copy this binary between containers in ECS at runtime that may not include libc6-compat.

## How I've tested this PR:
* Ran `go test ./... -run 'TestBasic/secure:_true' -v -p 1 -timeout 20m`
* Acceptance tests on this branch: https://github.com/hashicorp/terraform-aws-consul-ecs/compare/pglass/test-cgo-fix

👀 

## Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -86,7 +86,9 @@ jobs:
           GOOS: ${{ matrix.goos }}
           GOARCH: ${{ matrix.goarch }}
           LDFLAGS: ${{ needs.get-product-version.outputs.ldflags }}
+          CGO_ENABLED: "0"
         run: |
+          go env
           mkdir dist out
           go build -ldflags="$LDFLAGS" -o dist/ .
           zip -r -j out/${{ env.PKG_NAME }}_${{ needs.get-product-version.outputs.product-version }}_${{ matrix.goos }}_${{ matrix.goarch }}.zip dist/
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -86,7 +86,9 @@ jobs:
           GOOS: ${{ matrix.goos }}
           GOARCH: ${{ matrix.goarch }}
           LDFLAGS: ${{ needs.get-product-version.outputs.ldflags }}
+          CGO_ENABLED: "0"
         run: |
+          go env
```
cthain, line 91: Is the idea to leave this `go env` cmd in here; or, was it for testing?

cthain (approved):

.github/workflows/build.yml
pglass: I was going to keep it for future reference.

Actually, the [previous step](https://github.com/hashicorp/consul-ecs/runs/6599745996?check_suite_focus=true#step:3:16) installs Go and runs `go env` which is how I was able to check CGO was enabled. But I didn't see how to set CGO_ENABLED=0 persistently in the `actions/setup-go@v2` Github action, so I did it as part of the build and also run `go env` again 🤷

PR #97, opened 2022-05-26
Enforce lowercase service names

pglass:
With the auth method, the service token relies on a service identity, which requires a lowercase name. This updates the binary to:

- Validate service names are lowercase in the schema.json. Actually, require they match the [service identity regex](https://github.com/hashicorp/consul/blob/93b164aac37f83863707f05284b868dc8ea57719/acl/validation.go#L11). I also handled this for the gateway service name and update schema validation for this.
- When no service name is specified, fall back to the lowercase family name.

## How I've tested this PR:
Unit and acceptance tests

:eyes:

## Checklist:
- [x] Tests added
- [ ] CHANGELOG entry added

### DIFF
--- a/config/schema.json
+++ b/config/schema.json
@@ -62,7 +62,8 @@
       "properties": {
         "name": {
           "description": "The name the service will be registered as in Consul. Defaults to the Task family name if empty or null.",
-          "type": ["string", "null"]
+          "type": ["string", "null"],
+          "pattern": "(^$)|(^[a-z0-9]([a-z0-9-_]*[a-z0-9])?$)"
         },
         "tags": {
           "description": "List of string values that can be used to add service-level labels.",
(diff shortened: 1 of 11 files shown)


config/schema.json
```diff
@@ -62,7 +62,8 @@
       "properties": {
         "name": {
           "description": "The name the service will be registered as in Consul. Defaults to the Task family name if empty or null.",
-          "type": ["string", "null"]
+          "type": ["string", "null"],
+          "pattern": "(^$)|(^[a-z0-9]([a-z0-9-_]*[a-z0-9])?$)"
```
pglass, line 66: This is empty string `(^$)` or the [service identity regex](https://github.com/hashicorp/consul/blob/93b164aac37f83863707f05284b868dc8ea57719/acl/validation.go#L11) from Consul.

config/schema.json
```diff
@@ -357,37 +358,38 @@
         },
         "lanAddress": {
           "description": "LAN address and port for the gateway. If not specified, defaults to the task/node address.",
-          "type": "object",
+          "type": ["object", "null"],
```
cthain, line 361: Question: I'm not clear on why `null` is defined as a valid type when the fields themselves aren't required. Is this to allow for the possibility to define the config like `"lanAddress": null`?

cthain (approved):

config/schema.json
pglass: Yeah, it's kind of silly, but I setup the config schema this way because it slightly simplifies passing fields from the Terraform module. We don't have to worry in Terraform whether a field is `null` or typed (string, list, int) for it to pass validation, since there are so many fields we don't specify by default.

It may not make as much sense for some of these fields. And we've stopped using null so much for default values in the Terraform module, so maybe we could rethink this.

PR #103, opened 2022-06-08
Mesh gateway support when ACLs are enabled

cthain:
- Update the `acl-controller` to use the Consul server agent config to detect if mesh gateway WAN federation is enabled and update the anonymous token so that mesh gateways can communicate across datacenters.
- Add unit tests.

## How I've tested this PR:
- Unit tests
- Manually against the mesh-gateway example

:eyes: 

## Checklist:
- [x] Tests added
- [ ] CHANGELOG entry added

### DIFF
--- a/subcommand/acl-controller/command.go
+++ b/subcommand/acl-controller/command.go
@@ -1,12 +1,14 @@
 package aclcontroller
 
 import (
+	"bytes"
 	"context"
 	"flag"
 	"fmt"
 	"os"
 	"strings"
 	"sync"
+	"text/template"
 
 	"github.com/aws/aws-sdk-go/service/ecs"
 	"github.com/hashicorp/consul-ecs/awsutil"
@@ -15,6 +17,7 @@ import (
 	"github.com/hashicorp/consul/api"
 	"github.com/hashicorp/go-hclog"
 	"github.com/mitchellh/cli"
+	"github.com/mitchellh/mapstructure"
 )
 
 const (
@@ -134,6 +137,8 @@ func (c *Command) Help() string {
 // upsertConsulResources creates the necessary resources in Consul if they do not exist.
 // This includes the partition, client role and policy, client and service token auth methods,
 // and the necessary binding rules.
+// If mesh federation via mesh gateways is enabled the anonymous token will be updated with the
+// necessary read permissions.
 func (c *Command) upsertConsulResources(consulClient *api.Client, ecsMeta awsutil.ECSTaskMeta) error {
 	account, err := ecsMeta.AccountID()
 	if err != nil {
@@ -221,6 +226,9 @@ func (c *Command) upsertConsulResources(consulClient *api.Client, ecsMeta awsuti
 		return err
 	}
 	if err := c.upsertBindingRule(consulClient, serviceBindingRule); err != nil {
+		return err
+	}
+	if err := c.upsertAnonymousTokenPolicy(consulClient); err != nil {
 		return err
 	}
 	return nil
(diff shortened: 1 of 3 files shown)


subcommand/acl-controller/command.go
```diff
@@ -29,6 +32,10 @@ const (
 	// Binding rules don't support a '/' character, so we need compatible IAM role tag names.
 	authMethodServiceNameTag = "consul.hashicorp.com.service-name"
 	authMethodNamespaceTag   = "consul.hashicorp.com.namespace"
+
+	// anonTokenID is the well-known ID for the anonymous ACL token.
+	anonTokenID    = "00000000-0000-0000-0000-000000000002"
+	anonPolicyName = "anonymous-token-policy"
```
pglass, line 38: This is the same name [consul-k8s uses](https://github.com/hashicorp/consul-k8s/blob/aa05d3db2158e0887f49a0f5354893603ff6b6f2/control-plane/subcommand/server-acl-init/anonymous_token.go#L9-L43) 🤔 

Thinking about how this works if we have both and ECS an EKS dc/partition. Both sides would attempt to create and attach this policy, but that should be fine. Only one policy would be created, and both configure the policy with the same rules.

cthain: Yeah I intentionally used the same policy name as consul-k8s.. and I think this is OK because both consul-k8s and consul-ecs check if the policy already exists before attempting to create it. I think that the only time that both sides would attempt to create and attach the policy is in a race condition where they both read a non-existent policy at the same time, which should be rare.

subcommand/acl-controller/command.go
```diff
@@ -479,3 +561,71 @@ func (c *Command) writeOptions() *api.WriteOptions {
...
+}
+
+func (c *Command) templateData() templateData {
+	return templateData{Enterprise: c.flagPartitionsEnabled}
+}
+
+func (c *Command) anonymousPolicyRules() (string, error) {
+	rules := `
+{{- if .Enterprise }}
+namespace_prefix "" {
```
pglass, line 603: Should this have the `partition_prefix ""` scope ([consul-k8s for reference](https://github.com/hashicorp/consul-k8s/blob/aa05d3db2158e0887f49a0f5354893603ff6b6f2/control-plane/subcommand/server-acl-init/rules.go#L118-L137))?

subcommand/acl-controller/command_test.go
```diff
@@ -421,3 +441,124 @@ func makeAuthMethod(principals interface{}) *api.ACLAuthMethod {
...
+						// value that we got back when reading the policy.
+						expPolicyID = obsAnonTokenPolicy.ID
+					}
+
+					// expect that the policy is now attached to the anonymous token.
+					expAnonToken.Policies = append(expAnonToken.Policies, &api.ACLTokenPolicyLink{
+						ID:   expPolicyID,
+						Name: anonPolicyName})
+				}
+				// remove fields that don't compare
```
pglass, line 552: Another way we've ignored fields in comparisons is with [`cmpopts.IgnoreFields`](https://github.com/hashicorp/consul-ecs/blob/b1bd110b128f62bd136a453c75a4faa1ef4df97f/subcommand/mesh-init/command_test.go#L293-L298)

subcommand/acl-controller/command_test.go
```diff
@@ -421,3 +441,124 @@ func makeAuthMethod(principals interface{}) *api.ACLAuthMethod {
...
+				require.NoError(t, err)
+
+				if len(c.expPolicy) > 0 {
+					// if we expect the policy to be created then read it back to make sure
+					// it was and that it matches the expected
+					obsAnonTokenPolicy, _, err := consulClient.ACL().PolicyReadByName(anonPolicyName, &api.QueryOptions{})
+					require.NoError(t, err)
+					require.Equal(t, c.expPolicy, obsAnonTokenPolicy.Rules)
+
+					if expPolicyID == "" {
```
pglass, line 541: nit: To simplify this slightly, we could ignore checking the specific policy id, which would allow us to remove this `expPolicyId` variable. It should be okay to just check the policy name, since Consul doesn't let you create two policies with the same name (in the same partition/namespace).

Not a blocker.

subcommand/acl-controller/command.go
```diff
@@ -466,6 +488,66 @@ func (c *Command) upsertBindingRule(consulClient *api.Client, bindingRule *api.A
...
+// service:read and node:read.
+// The anonymous token is global so it is replicated from the primary DC to all secondary
+// DCs, which is why we only update it if this is the primary datacenter.
+func (c *Command) upsertAnonymousTokenPolicy(consulClient *api.Client, agentConfig AgentConfig) error {
+	consulDC, primaryDC, err := c.consulDatacenterList(agentConfig)
+	if err != nil {
+		return fmt.Errorf("failed to list Consul datacenters: %w", err)
+	}
+
+	if !agentConfig.DebugConfig.MeshGatewayWANFederationEnabled || consulDC != primaryDC {
```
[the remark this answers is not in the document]
cthain: Yes, it would also be helpful to have a field that indicates whether or not the agent is running an enterprise version. There was this thread that mentioned automation tooling and Nomad are parsing this info out of the version string.. but that seems a bit dubious and broke in 1.12.0. https://hashicorp.slack.com/archives/C045NJD17/p1650927945635299

I guess the best place to bring this up would be in #team-consul, thoughts?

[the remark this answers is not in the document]
cthain: Yes, it would also be helpful to have a field that indicates whether or not the agent is running an enterprise version. There was this thread that mentioned automation tooling and Nomad are parsing this info out of the version string.. but that seems a bit dubious and broke in 1.12.0. https://hashicorp.slack.com/archives/C045NJD17/p1650927945635299

I guess the best place to bring this up would be in #team-consul, thoughts?

subcommand/acl-controller/command.go
```diff
@@ -466,6 +488,66 @@ func (c *Command) upsertBindingRule(consulClient *api.Client, bindingRule *api.A
...
+		return nil
+	}
+
+	c.log.Info("Configuring anonymous token", "datacenter", consulDC, "primary-datacenter", primaryDC)
+
+	policy, _, err := consulClient.ACL().PolicyReadByName(anonPolicyName, &api.QueryOptions{})
```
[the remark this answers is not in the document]
cthain: Good call, will do.

subcommand/acl-controller/command_test.go
cthain: OK, this suggestion is fine with me. This is how I originally had it but thought that checking the ID would be more thorough. But admittedly this check is kind of janky so I'm fine to remove it.

subcommand/acl-controller/command.go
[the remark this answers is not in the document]
pglass: > I guess the best place to bring this up would be in #team-consul, thoughts?

Yep!

PR #105, opened 2022-06-08
Fix namespace reconciliation error

pglass:
This fixes an error message in the acl-controller where it was attempting to use the empty string for the namespace when creating a namespace. When this error occurred while reconciling namespaces, the controller would abort reconciliation, which means it would not proceed to cleanup ACL tokens.

To address that, this also changes the controller to attempt to reconcile resources even if a namespace reconciliation error occurred.

## How I've tested this PR:
Unit tests

👀 

## Checklist:
- [x] Tests added
- [x] CHANGELOG entry added

### DIFF
--- a/controller/controller.go
+++ b/controller/controller.go
@@ -6,6 +6,7 @@ import (
 	"time"
 
 	"github.com/hashicorp/go-hclog"
+	"github.com/hashicorp/go-multierror"
 )
 
 const DefaultPollingInterval = 10 * time.Second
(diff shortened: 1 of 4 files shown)


controller/controller.go
```diff
@@ -45,17 +46,20 @@ func (c *Controller) reconcile() error {
 		return fmt.Errorf("listing resources: %w", err)
 	}
 
+	var merr error
 	if err = c.Resources.ReconcileNamespaces(resources); err != nil {
-		return fmt.Errorf("reconciling namespaces: %w", err)
+		c.Log.Error("error reconciling namespaces", "err", err)
```
cthain, line 51: I'm wondering if the log messages here are necessary since there is also a log message in `ctrl.reconcile()` :thinking: ?.. Any errors here will get logged twice.

cthain (approved):

controller/controller.go
pglass: I guess I was thinking I should log the error immediately, but now that I'm seeing this again, you're right. I'll remove these log lines!

PR #113, opened 2022-07-06
Fix anonymous token policy description

cthain:
- Change the `Description` of the anonymous token policy so that it is exactly the same as the one created by the `consul-k8s` ACL controller.
- Addressed dependabot security alert for `Masterminds/goutils` module

## How I've tested this PR:
- Updated unit test

:eyes: 

## Checklist:
- [x] Tests added
- [x] CHANGELOG entry added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+## Unreleased
+
+BUG FIXES:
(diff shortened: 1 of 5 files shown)


pglass (approved):
👍 thanks! Be sure to backport this to the release/0.5.x branch so we can do a patch release.

PR #115, opened 2022-07-14
Use Login API instead of the CLI

pglass:
Update mesh-init to use [hashicorp/consul-awsauth](https://github.com/hashicorp/consul-awsauth) and the Login API instead of the `consul login` CLI command.

- Remove the `consulLogin.extraLoginFields` config option. This was a list of additional flags for `consul login`, which is no longer used.
  - Previously, the `-partition <partition>` flag needed to be passed to `extraLoginFields`. Since that has been removed, the partition for the login request is copied from the service or gateway config (`service.partition` or `gateway.partition` field if non-empty). This should be fine since the service and auth method instance always exist in the same partition where the service is registering.
- Add the following `consulLogin` fields:
  - `meta`: equivalent to the `-meta` option
  - `region`: to force an AWS region. By default, we'll discover the region from task metadata so this will be rarely used, but is useful for unittests.
  - `stsEndpoint`: to use a particular AWS STS endpoint URL (passed along to consul-awsauth)
  - `serverIdHeaderValue`: a server id header for an additional form of request validation (passed along to consul-awsauth)

## How I've tested this PR:
- [x] Unit tests
- [x] Acceptance tests in https://github.com/hashicorp/terraform-aws-consul-ecs/commit/e4260600b53d50ae894646c95e238dce0796a914

:eyes:

## Checklist:
- [x] Tests added
- [x] CHANGELOG entry added


erichaberkorn (approved):

erichaberkorn (commented):
Could you add a changelog entry for this? It is a breaking change because it removes the `ExtraLoginFlags` field.

pglass:
I updated this to handle mesh gateway as well, and added the changelog

PR #116, opened 2022-08-18
set PRODUCT_VERSION for default docker build

alvin-huang:

In `actions-docker-build` we [pass](https://github.com/hashicorp/actions-docker-build/blob/05c370a26e61b06be46c5095d6e914c9f0ea4f3d/scripts/docker_build#L49) `PRODUCT_VERSION` to the docker build command. Since this was not set, the label did not populate properly which is used in a comparison to determine the `minor-latest` and `latest` docker image tags. 

```
docker inspect hashicorp/consul-ecs:latest -f={{json '.Config.Labels'}} | jq 

{
  "description": "consul-ecs provides first-class integration between Consul and AWS ECS.",
  "maintainer": "Consul Team <consul@hashicorp.com>",
  "name": "consul-ecs",
  "org.opencontainers.image.authors": "Consul Team <consul@hashicorp.com>",
  "org.opencontainers.image.description": "consul-ecs provides first-class integration between Consul and AWS ECS.",
  "org.opencontainers.image.documentation": "https://www.consul.io/docs/ecs",
  "org.opencontainers.image.source": "https://github.com/hashicorp/consul-ecs",
  "org.opencontainers.image.title": "consul-ecs",
  "org.opencontainers.image.url": "https://www.consul.io/",
  "org.opencontainers.image.vendor": "HashiCorp",
  "org.opencontainers.image.version": "",
  "release": "",
  "summary": "consul-ecs provides first-class integration between Consul and AWS ECS.",
  "vendor": "HashiCorp",
  "version": "" <---- needs to be set
}

```

How I've tested this PR:
 - build the image up to the point of label creation and pass in `--build-arg PRODUCT_VERSION=1.2.3`
 - inspect the image for the label with the above command

How I expect reviewers to test this PR:
- same as above

## Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added


Dockerfile
```diff
@@ -7,28 +7,28 @@
...
 ARG TARGETOS TARGETARCH
 # Export BIN_NAME for the CMD below, it can't see ARGs directly.
 ENV BIN_NAME=$BIN_NAME
-ENV VERSION=$VERSION
+ENV VERSION=$PRODUCT_VERSION
 
 LABEL description="consul-ecs provides first-class integration between Consul and AWS ECS." \
       maintainer="Consul Team <consul@hashicorp.com>" \
       name=$BIN_NAME \
       release=$VERSION \
```
cthain, line 20: The value of `release` will get set to the same value as `$PRODUCT_VERSION` because it will grab the value from the `$VERSION` env var... but for consistency we should probably use `$PRODUCT_VERSION` directly instead of transitively.

```suggestion
      release=$PRODUCT_VERSION \
```

cthain (commented):
One minor suggestion, otherwise LGTM.

alvin-huang:
Good catch, that makes sense to me!

cthain (approved):

PR #118, opened 2022-11-15
Add proxy.publicListenerPort config option

pglass:

This adds a `proxy.publicListenerPort` config field to allow configuring the Envoy public listener port. This defaults to 20000, which was the hard-coded value.

## How I've tested this PR:

* Unit tests
* Pending acceptance tests in 

## Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added


config/resources/test_config_null_nested_fields.json
```diff
@@ -67,6 +67,7 @@
   },
   "proxy": {
     "config": null,
+    "publicListenerPort": 0,
```
cthain, line 70: This is a `null`able type according to the schema. Do we want to test `null` here instead of `0`?

config/types.go
```diff
@@ -249,11 +253,18 @@ func (a *AgentServiceConnectProxyConfig) ToConsulType() *api.AgentServiceConnect
 	return result
 }
 
+func (a *AgentServiceConnectProxyConfig) GetPrimaryListenerPort() int {
```
cthain, line 256: Just to be consisent with the naming.
```suggestion
func (a *AgentServiceConnectProxyConfig) GetPublicListenerPort() int {
```

cthain (approved):
Looks good. Just a minor comment and question from me.

PR #70, opened 2022-04-15
Extract unit test utils into testutils package

pglass:
## Changes proposed in this PR:
Add a `testutils` package that extracts common unit test behaviors and encapsulates their setup/cleanup.

- Starting a Consul test server
- Starting an ECS task metadata server
- Creating a temp dir
- Setting CONSUL_ECS_CONFIG_JSON from a `config.Config`

## How I've tested this PR:
- Unit tests

## How I expect reviewers to test this PR:
👀 

## Checklist:
- [x] Tests added
- [x] ~CHANGELOG entry added~ n/a

    [HashiCorp engineers only. Community PRs should not add a changelog entry.]::
    [Changelog entries should use present tense, e.g. "Add support for..."]::

PR #120, opened 2022-11-16
Enable CRT for 0.4.x

pglass:

This backports the CRT workflow to the newly-created `release/0.4.x` branch, so that we can do future patch releases of 0.4.x.

Easier to walk through commit by commit:

* d06c3dc38e907acbd414964478f06a19275b95cd - Copy the CRT config as-is from main
* 992a2ea010d4ee00065e4ef582d7cb265d5beba1 - Remove the Circle CI config and scripts in ./build-support
* 4a637941b7861e9469ef70f9c4c8faf6bd041eea - Test with Consul 1.11.11 in CI since 1.11.x is the supported Consul version for 0.4.x. This corrects the CRT config to include release/0.4.x as a release branch, and sets current version to 0.4.1-dev.
* a58e9fb2af7cad5c97e01625055c1347456bcd49 - Add the missing .go-version file that I forgot to copy from main
* b5c9be228eedc6ae11d3086ab495622f7d4d8855 - Correct the .dockerignore that I forgot to copy from main
* 4083b2e12b1be8786b25be2000f5107f22670a0f - Backport a testutil refactor. Actually in 0.4.1, the Circle CI job wasn't running the enterprise tests because it was not passing a `--tags=enterprise` option to `go test`, so one enterprise test case was broken. I fixed this issue before in #70, and since that PR contained only test changes I backported the whole thing (minus the Circle CI config) to reduce the diff with main and 0.5.x
* 4d45cdb84c4be6024ea7d571633842f6a7722322 - In 0.5 with the switch to CRT, we changed the version format to `v0.Y.X` instead of `0.Y.X` (lack of the leading `v`). This caused a the version smoke test check to fail in the CRT workflow, so I kept the `consul-ecs --version` output as-is in 0.4.x (no leading `v`), and updated the CRT workflow to match.

## How I've tested this PR:

* CI pipeline + ran tests locally with Consul 1.11.11

👀 

## Checklist:
- [x] Tests added
- [ ] CHANGELOG entry added


.dockerignore
```diff
@@ -1,3 +1,3 @@
 *
 */*
-!pkg/bin/linux_amd64/consul-ecs
```
cthain, line 3: this line still exists on `main`.. but I don't think that it's required since all the outputs go in to the `dist/` directory. Not an issue here but maybe we should remove this from `main`?

cthain (approved):
Thanks for the detailed commit breakdowns. Made the review much easier!

I think pulling in the utility test updates is a really good idea.

.dockerignore
pglass: I addressed your other comment, here and also tacked on a commit in #118: https://github.com/hashicorp/consul-ecs/pull/118/commits/11b126e7d495645a97639b5a9008ff97d8e15d3c, which will go into main and then will get backported to 0.5.x

PR #124, opened 2022-11-17
Update to Go 1.19

pglass:

This updates to Go 1.19. This requires fixes a couple of things to satisfy golangci-lint:

* Re-run `go fmt`, which is a noisy change because Go 1.19 formats comments.
* Stop using ioutil which is deprecated
* Also, update golangci-lint which I think is necessary for Go 1.19 because I got an error

Also, piggybacks a bump to Consul 1.12.6 (instead of 1.12.0)

## How I've tested this PR:

CI tests + local tests + local security scan

## Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added


.github/workflows/bin-ci.yml
```diff
@@ -4,22 +4,38 @@ on:
...
+      go-version: ${{ steps.get-go-version.outputs.go-version }}
+    steps:
+      - uses: actions/checkout@v2
+      - name: Determine Go version
+        id: get-go-version
+        # We use .go-version as our source of truth for current Go
+        # version, because "goenv" can react to it automatically.
+        run: |
+          echo "Building with Go $(cat .go-version)"
+          echo "::set-output name=go-version::$(cat .go-version)"
```
pglass, line 23: I copied this over from the build.yml to read go version from `.go-version` instead of using an environment variable.

cthain: Ya, this is way better than having it specified in two places.

cthain (approved):
Nice, thanks for doing this. +1 for bumping the Consul version. Reviewing with whitespace changes suppressed wasn't bad at all.

PR #133, opened 2023-03-03
Support Consul 1.15.0

cthain:
Some of the APIs in Consul 1.15 have been updated to return a `nil` error and an empty resource in the case of a not found scenario. Previous versions of Consul returned an error containing the string `403 (ACL not found)`. These difference caused startup failures in the `acl-controller`.

This PR adds support for Consul 1.15 while maintaining compatibility with the existing supported Consul versions.

## How I've tested this PR:
Unit tests

:eyes: 

## Checklist:
- [x] Tests updated
- [x] CHANGELOG entry added

### DIFF
--- a/controller/resource_test.go
+++ b/controller/resource_test.go
@@ -345,7 +345,6 @@ func TestTaskStateReconcile(t *testing.T) {
 					Partition: exp.Partition,
 				})
 				require.Error(t, err)
-				require.Contains(t, err.Error(), "403 (ACL not found)")
 				require.Nil(t, tok)
 			}
 			for _, exp := range c.expExist {
@@ -473,7 +472,6 @@ func TestReconcileNamespaces(t *testing.T) {
 					}
 				}
 			} else {
-				require.Error(t, err)
 				require.Nil(t, xnsPolicy)
 			}
 		})
--- a/controller/resource_test.go
+++ b/controller/resource_test.go
@@ -345,6 +345,7 @@ func TestTaskStateReconcile(t *testing.T) {
 					Partition: exp.Partition,
 				})
 				require.Error(t, err)
+				require.Contains(t, err.Error(), "ACL not found")
 				require.Nil(t, tok)
 			}
 			for _, exp := range c.expExist {
--- a/.github/workflows/bin-ci.yml
+++ b/.github/workflows/bin-ci.yml
@@ -53,12 +53,12 @@ jobs:
     strategy:
       matrix:
         consul-version:
-        - 1.12.7
-        - 1.12.7+ent
-        - 1.13.4
-        - 1.13.4+ent
-        - 1.14.2
-        - 1.14.2+ent
+        - 1.13.6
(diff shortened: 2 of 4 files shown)


controller/resource_test.go
```diff
@@ -473,7 +472,6 @@ func TestReconcileNamespaces(t *testing.T) {
 					}
 				}
 			} else {
-				require.Error(t, err)
```
cthain, line 476: For simplicity, I decided to just remove the test for the a non-nil error here. The other option would have been to check the version of the Consul client and selectively test for the error if the version is < 1.15.0. This seemed a bit heavy given the scope of the changes required to support 1.15.

controller/resource_test.go
```diff
@@ -345,7 +345,6 @@ func TestTaskStateReconcile(t *testing.T) {
 					Partition: exp.Partition,
 				})
 				require.Error(t, err)
-				require.Contains(t, err.Error(), "403 (ACL not found)")
```
cthain, line 348: This error string is only applicable to Consul versions prior to 1.15 so I decided to just take it out.

pglass: minor: I think all both pre-1.15 and post-1.15 versions of Consul contain `ACL not found`, just in a different parts of the error message? It would be good to still check we are getting the correct type of error.

controller/resource_test.go
pglass: Are we getting a nil error here because of a "bug" in the `api.Client` with 1.15? Maybe comment this out instead and add a TODO to circle back when the api.Client is fixed.

pglass (approved):
👍 A couple of minor thoughts

pglass:
Oh, should we update CI to test with 1.15.0 ?

roncodingenthusiast (approved):
Code looks good to me, no additional comment. +1 on the comment from Paul checking the correct error type.

controller/resource_test.go
cthain: I don't think that this is a "bug" because it was intentionally changed to return `404`. In which case the client consumes the `404` returns a nil error and nil resournce to indicate "not found":
1. https://github.com/hashicorp/consul/blob/main/agent/acl_endpoint.go#L169-L174
1. https://github.com/hashicorp/consul/blob/main/api/acl.go#L998-L1001
1. https://github.com/hashicorp/consul/blob/main/api/api.go#L1243-L1244

Previously this resolved to a `403`: https://github.com/hashicorp/consul/blob/v1.14.4/agent/acl_endpoint.go#L157-L160, so a non nil error was returned by the client.

controller/resource_test.go
cthain: Great catch Paul. You're correct that both versions have `ACL not found`. I've updated the test to re-instate the check.

controller/resource_test.go
pglass: Ah. That makes sense

pglass (approved):

PR #141, opened 2023-05-23
Add FIPS build

skpratt:
Add FIPS versions of consul-ecs

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -186,7 +186,7 @@ jobs:
             public.ecr.aws/hashicorp/${{ env.repo }}-fips:${{ env.version }}
           dev_tags: |
             hashicorppreview/${{ env.repo }}-fips:${{ env.version }}
-            docker.io/hashicorppreview/${{ env.repo }}-fips:${{ env.version }}${{ github.sha }}
+            docker.io/hashicorppreview/${{ env.repo }}-fips:${{ env.version }}-${{ github.sha }}
 
   upload-dev-docker:
     name: Upload dev image (hashicorpdev/consul-ecs:<commit>)
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -21,7 +21,7 @@ jobs:
         # version, because "goenv" can react to it automatically.
         run: |
           echo "Building with Go $(cat .go-version)"
-          echo "::set-output name=go-version::$(cat .go-version)"
+          echo "go-version=$(cat .go-version)" >> $GITHUB_OUTPUT
 
   get-product-version:
     runs-on: ubuntu-latest
@@ -35,9 +35,9 @@ jobs:
         id: get-product-version
         run: |
           make version
-          echo "::set-output name=product-version::$(make version)"
-          echo "::set-output name=git-short-sha::$(git rev-parse --short HEAD)"
-          echo "::set-output name=ldflags::-s -w -X github.com/hashicorp/consul-ecs/version.GitCommit=$(git rev-parse --short HEAD)"
+          echo "product-version=$(make version)" >> $GITHUB_OUTPUT
+          echo "git-short-sha=$(git rev-parse --short HEAD)" >> $GITHUB_OUTPUT
+          echo "ldflags=-s -w -X github.com/hashicorp/consul-ecs/version.GitCommit=$(git rev-parse --short HEAD)" >> $GITHUB_OUTPUT
 
   generate-metadata-file:
     needs: get-product-version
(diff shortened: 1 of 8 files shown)


.github/workflows/build.yml
```diff
@@ -90,17 +95,17 @@ jobs:
...
           fi
 
       - uses: actions/upload-artifact@0b7f8abb1508181956e8e162db84b466c27e18ce # v3.1.2
         with:
-          name: ${{ env.PKG_NAME }}_${{ needs.get-product-version.outputs.product-version }}_${{ matrix.goos }}_${{ matrix.goarch }}.zip
-          path: out/${{ env.PKG_NAME }}_${{ needs.get-product-version.outputs.product-version }}_${{ matrix.goos }}_${{ matrix.goarch }}.zip
+          name: ${{ env.PKG_NAME }}_${{ needs.get-product-version.outputs.product-version }}${{ matrix.fips }}_${{ matrix.goos }}_${{ matrix.goarch }}.zip
+          path: out/${{ env.PKG_NAME }}_${{ needs.get-product-version.outputs.product-version }}${{ matrix.fips }}_${{ matrix.goos }}_${{ matrix.goarch }}.zip
 
   build-docker-default:
```
pglass, line 110: Are we going to build FIPS docker images?

.github/workflows/build.yml
```diff
@@ -134,35 +142,51 @@ jobs:
...
+          - { arch: "arm64" }
+          - { arch: "amd64" }
+    env:
+      repo: ${{ github.event.repository.name }}
+      version: ${{ needs.get-product-version.outputs.product-version }}+fips1402
+
+    steps:
+      - uses: actions/checkout@8e5e7e5ab8b370d6c329ec480221332ada57f0ab # v3.5.2
+      - name: Docker FIPS Build (Action)
+        uses: hashicorp/actions-docker-build@v1
```
cthain, line 170: The version for `actions-docker-build` should be pinned to a SHA.

.github/workflows/build.yml
```diff
@@ -134,35 +142,51 @@ jobs:
...
+            echo "Test PASSED"
+          version: ${{ env.version }}
+          target: release-fips-default
+          arch: ${{ matrix.arch }}
+          tags: |
+            docker.io/hashicorp/${{ env.repo }}-fips:${{ env.version }}
+            public.ecr.aws/hashicorp/${{ env.repo }}-fips:${{ env.version }}
+          dev_tags: |
+            hashicorppreview/${{ env.repo }}-fips:${{ env.version }}
+            docker.io/hashicorppreview/${{ env.repo }}-fips:${{ env.version }}${{ github.sha }}
```
cthain, line 189: I think we're missing a `-` between the version and the SHA:

```suggestion
            docker.io/hashicorppreview/${{ env.repo }}-fips:${{ env.version }}-${{ github.sha }}
```

Makefile
```diff
@@ -28,6 +28,10 @@ dev: dist
 	GOARCH=$(ARCH) GOOS=$(OS) go build -ldflags "$(LD_FLAGS)" -o $(BIN)
 .PHONY: dev
 
+fips-dev: dist
```
cthain, line 31: What do you think about renaming this target `dev-fips`? To me it feels a bit more consistent with the other target naming... `make dev-fips` feels a bit more natural to me than `make fips-dev`. Below we have `make docker-fips`, which is consistent with this.

```shell
$ make dev-fips
$ make docker-fips
```

Not a blocker at all. I'll leave it up to you.

version/fips_build.go
```diff
@@ -0,0 +1,27 @@
+//go:build fips
```
cthain, line 1: Could we name the tag-dependent build files as follows?:

```
version/build.go
version/build_fips.go
```

subcommand/version/command.go
```diff
@@ -12,6 +12,7 @@ import (
 type Command struct {
 	UI      cli.Ui
 	Version string
+	FIPS    string
```
cthain, line 15: Are we intending to output this in the `consul-ecs version` command? I don't think that it will actually get reported since we only report `c.Version` in the CLI output: https://github.com/hashicorp/consul-ecs/blob/7b44fa3510d07f2874bff13b0b4df3cf8aae6703/subcommand/version/command.go#L19

cthain (commented):
This is looking good. I just had a couple minor questions/comments.

.github/workflows/build.yml
skpratt: I think the SHA pinning is only for external actions, that way CRT can more gracefully make small changes with our homegrown actions - at least that is how it appears to be in the other Consul repos

Makefile
skpratt: done

subcommand/version/command.go
skpratt: Removed - thank you for catching. For consistency wrt to smoke tests and user scripts, and given that this is OSS so the FIPS build bits are available for inspection, we aren't changing the version format to verbose multi-field/multi-line in k8s, dataplane, nor ECS.

cthain: :ok_hand: Thanks for the explanation.

cthain (approved):
:+1:

PR #143, opened 2023-06-22
Added support for server discovery, TLS and ACL login

Ganeshrockz:

This PR introduces the following changes to consul ECS

- Adds consul server connection manager discover the consul server agents from the ECS control plane in the new architecture
- Adds support for ACL login in the Control plane module
- Adds support for protocol specific TLS (HTTPS & gRPC) configurations
- Adds the `consulServers` field to `schema.json`. The top level field looks like
```json
{
   "consulServers": {
          "hosts": "<addresses>",
          "https": true, // <Defaults to true>, 
          "tls": true, // <Defaults to true>,
          "httpPort":  8501, //<Defaults to 8501 since https is enabled by default>
          "grpcPort": 8503, //<Defaults to 8503 since tls is enabled by default>
          "caCertFile": "",
          "httpsCACertFile": "",
          "skipServerWatch": false,
          "tlsServerName": ""
    }
}
```
- Configures the new client which can talk to the server directly. 
- Stores the ACL token returned by the server connection manager in a separate file and writes it to a shared volume. This will be later used by the dataplane to configure itself.
- Server watch isn't implemented yet. Will be done in subsequent iterations.

## How I've tested this PR:

Unit tests for now. I have decided to split the agentless implementation into multiple PRs and have the `ganeshrockz/agentless` branch as the target branch. Once all the necessary changes go into it, we need to rely on acceptance/integration tests before merging to main.

## Checklist:
- [X] Tests added
- [ ] CHANGELOG entry added (Not needed)


subcommand/mesh-init/command_test.go
```diff
@@ -146,15 +142,16 @@ func TestRun(t *testing.T) {
...
-		"auth method enabled": {
-			consulLogin: config.ConsulLogin{
-				Enabled:       true,
-				IncludeEntity: true,
-				Meta: map[string]string{
-					"unittest-tag": "12345",
-				},
-			},
-		},
+		// "auth method enabled": {
```
Ganeshrockz, line 145: This test needs to be revisited later on. In the current state, service registration fails stating that the token doesn't have a `service:write` permission.

subcommand/mesh-init/command.go
```diff
@@ -63,31 +60,68 @@ func (c *Command) Run(args []string) int {
...
+	defer cancel()
+
 	taskMeta, err := awsutil.ECSTaskMetadata()
 	if err != nil {
 		return err
 	}
 
 	cfg := api.DefaultConfig()
 
+	consulClient, err := api.NewClient(cfg)
```
Ganeshrockz, line 73: Keep the old version of the client intact so that unit tests don't fail in this branch. WIll eventually be removed when we start registering services with the new version of the client

config/config.go
```diff
@@ -0,0 +1,224 @@
...
+	return cfg, nil
+}
+
+func (c *Config) getNamespace() string {
+	namespace := c.Service.Namespace
+	if namespace == "" && c.Gateway != nil {
+		namespace = c.Gateway.Namespace
+	}
+
+	return namespace
```
pglass, line 135: minor: With these `getNamespace()` and `getPartition()` functions, it looks possible to accidentally use the namespace from the `service` config but the partition from the `gateway` config (if I've misconfigured things).

I'd recommend we add an `isGateway()` function that we can use here (and elsewhere)

```go
func (c *Config) IsGateway() bool {
	return c.Gateway != nil && c.Gateway.Kind != ""
}
```

And then this becomes:

```suggestion
	if c.IsGateway() {
		return c.Gateway.Namespace
	}
	return c.Service.Namespace
```

config/config_test.go
```diff
@@ -0,0 +1,333 @@
...
+					c.cfg.Service.Namespace = "test-namespace"
+					c.cfg.Service.Partition = "test-partition"
+				}
+
+				// Enable ACLs to test with the auth method
+				srvConfig = testutil.ConsulACLConfigFn
+				_, apiCfg := testutil.ConsulServer(t, srvConfig)
+				consulClient, err := api.NewClient(apiCfg)
+				require.NoError(t, err)
+				fakeAws := testutil.AuthMethodInit(t, consulClient, c.taskMeta.Family, c.cfg.ConsulLogin.Method)
```
pglass, line 137: In this test, we should always create the auth method in the default namespace, regardless of where the namespace service/task is registering into (this is what the ACL controller does).

config/config_test.go
```diff
@@ -0,0 +1,333 @@
...
+		})
+	}
+}
+
+func TestConsulServerConnManagerConfig_TLS(t *testing.T) {
+	caFile, err := os.CreateTemp("", "")
+	t.Cleanup(func() {
+		_ = os.RemoveAll(caFile.Name())
+	})
+	require.NoError(t, err)
```
pglass, line 172: nit: check for error before adding the cleanup

```suggestion
	require.NoError(t, err)
	t.Cleanup(func() {
		_ = os.RemoveAll(caFile.Name())
	})
```

config/config_test.go
```diff
@@ -0,0 +1,333 @@
...
+					Addresses: "consul.dc1.address",
+					GRPCPort:  8503,
+				}
+			},
+		},
+		"basic flags without TLS or ACLs with custom gRPC port": {
+			cfg: &Config{
+				ConsulServers: ConsulServers{
+					Hosts:    "consul.dc1.address",
+					GRPCPort: 8502,
```
pglass, line 59: This test case (`basic flags without TLS or ACLs with custom gRPC port`) looks the same as the one above (`basic flags without TLS or ACLs`) because a gRPC port is specified in both test cases. I think we could remove one of these test cases, and then use different gRPC ports in the test cases that remain. What do you think?

config/config_test.go
```diff
@@ -0,0 +1,333 @@
...
+				},
+			},
+			expConfig: func(t awsutil.ECSTaskMeta) discovery.Config {
+				return discovery.Config{
+					Addresses: "consul.dc1.address",
+					GRPCPort:  8502,
+				}
+			},
+		},
+		"default TLS": {
```
pglass, line 69: nit: naming - because we aren't testing the config parsing/validation here, these tests aren't validating what default values are used, so it makes sense to just name these tests based on which settings are enabled

```suggestion
		"TLS enabled": {
```

config/config_test.go
```diff
@@ -0,0 +1,333 @@
...
+}
+
+func TestClientConfig(t *testing.T) {
+	caFile, err := os.CreateTemp("", "")
+	t.Cleanup(func() {
+		_ = os.RemoveAll(caFile.Name())
+	})
+	require.NoError(t, err)
+	_, err = caFile.WriteString(testCA)
+	require.NoError(t, err)
```
pglass, line 229: minor: we write the testCA file in a couple of places, so we could create a helper function for it.

config/config_test.go
```diff
@@ -0,0 +1,333 @@
...
+				Scheme: "https",
+				TLSConfig: api.TLSConfig{
+					Address: "consul.dc1.address",
+					CAFile:  caFile.Name(),
+				},
+			},
+		},
+		"TLS with CaCertPEM": {
+			setupEnv: func() error {
+				return os.Setenv(consulCACertPemEnvVar, testCA)
```
pglass, line 299: minor: There is actually [`t.Setenv`](https://pkg.go.dev/testing#T.Setenv) we could use for this now

subcommand/mesh-init/command.go
pglass: nit: Maybe add a `TODO` comment to keep track of this

subcommand/mesh-init/command.go
```diff
@@ -63,31 +60,68 @@ func (c *Command) Run(args []string) int {
...
+	if err != nil {
+		return fmt.Errorf("constructing consul client: %s", err)
+	}
+
+	serverConnMgrCfg, err := c.config.ConsulServerConnMgrConfig(taskMeta)
+	if err != nil {
+		return fmt.Errorf("constructing server connection manager config: %s", err)
+	}
+	serverConnMgrCfg.ServerWatchDisabled = false
```
pglass, line 82: We'll need a config option for server watch disabled. Users will need to disable the server watch whenever their Consul servers are behind a load balancer (per [the docs](https://developer.hashicorp.com/consul/docs/connect/dataplane/consul-dataplane#server-watch-disabled))

subcommand/mesh-init/command.go
```diff
@@ -63,31 +60,68 @@ func (c *Command) Run(args []string) int {
...
+		return fmt.Errorf("unable to create consul server watcher: %s", err)
+	}
+
+	go watcher.Run()
+	defer watcher.Stop()
+
+	state, err := watcher.State()
+	if err != nil {
+		return fmt.Errorf("unable to fetch consul server watcher state: %s", err)
+	}
```
pglass, line 95: non-blocking / informational: For a future PR, because the unified consul-ecs-control-plane binary will be long running in agentless, we'll need to rebuild the Consul client whenever the library switches to another Consul server.

The library has a basic [Subscribe](https://pkg.go.dev/github.com/hashicorp/consul-server-connection-manager@v0.1.2/discovery#Watcher.Subscribe) function that we can use to receive updates about Consul server addresses, as one way to do this. (Alternatively, we can periodically call `watcher.State()` and detect a server change that way which is also fine).

subcommand/mesh-init/command.go
```diff
@@ -63,31 +60,68 @@ func (c *Command) Run(args []string) int {
...
+		// previous version of the client (that talks through agents)
+		// can pass.
 		cfg.TokenFile = tokenFile
 
-		// The just-created token is not immediately replicated to Consul server followers.
-		// Mitigate against this by waiting for the token in stale consistency mode.
-		if err := c.waitForTokenReplication(tokenFile); err != nil {
-			return err
-		}
+		c.log.Info("successfully logged in", "token-file", tokenFile)
```
pglass, line 111: minor: The server-connection-manager will [log about the ACL login](https://github.com/hashicorp/consul-server-connection-manager/blob/main/discovery/watcher.go#L445), so this log line could be changed to be more specific:

```suggestion
		c.log.Info("wrote ACL token to shared volume", "token-file", tokenFile)
```

subcommand/mesh-init/command.go
```diff
@@ -178,211 +212,6 @@ func (c *Command) realRun() error {
...
-	}
-
-	loginDataJson, err := json.Marshal(loginData)
-	if err != nil {
-		return "", err
-	}
-	return string(loginDataJson), err
-}
-
-func (c *Command) waitForTokenReplication(tokenFile string) error {
```
pglass, line 324: Note: I think it's fine to remove this for now, because of the [sleep hack](https://github.com/hashicorp/consul-server-connection-manager/blob/main/discovery/acl.go#L68-L73) in the connection manager lib. It may be worth adding a comment about this where the ACL login now happens, in case that sleep hack isn't enough to reliably prevent additional cases of this issue (which it likely isn't).

testutil/config.go
```diff
@@ -37,13 +40,13 @@ func SetECSConfigEnvVar(t *testing.T, val interface{}) {
 	require.NoError(t, err)
 
 	t.Cleanup(func() {
-		_ = os.Unsetenv(config.ConfigEnvironmentVariable)
+		_ = os.Unsetenv(configEnvVar)
 	})
```
pglass, line 44: nit: Another opportunity to use `t.Setenv`

subcommand/mesh-init/command.go
Ganeshrockz: Yeah. I've added a note in the PR description as well. Server watch will be taken up later on whenever we make the consul-ecs control plane a long running one.

config/config_test.go
Ganeshrockz: Sounds good

config/config.go
```diff
@@ -0,0 +1,224 @@
...
+			return discovery.Config{}, err
+		}
+
+		cfg.Credentials = credentials
+	}
+
+	return cfg, nil
+}
+
+func (c *Config) ClientConfig() *api.Config {
```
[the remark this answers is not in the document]
Ganeshrockz: Yup. This will be done in a later PR when we convert the control plane to become long running

config/config.go
```diff
@@ -0,0 +1,224 @@
...
+
+	return cfg
+}
+
+func (c *Config) getDiscoveryCredentials(taskMeta awsutil.ECSTaskMeta) (discovery.Credentials, error) {
+	cfg := discovery.Credentials{
+		Type: discovery.CredentialsTypeLogin,
+		Login: discovery.LoginCredential{
+			Datacenter: c.ConsulLogin.Datacenter,
+			Namespace:  c.getNamespace(),
```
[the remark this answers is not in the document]
Ganeshrockz: Does the same apply to partitions as well?

config/config_test.go
Ganeshrockz: Removed the custom partition and namespace for enterprise usecases

config/config.go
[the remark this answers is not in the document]
Ganeshrockz: I guess so after looking into the current ACL controller

config/config.go
```diff
@@ -0,0 +1,224 @@
...
+		caCert := os.Getenv(consulCACertPemEnvVar)
+		if caCert != "" {
+			cfg.TLSConfig.CAPem = []byte(caCert)
+		} else if c.ConsulServers.CACertFile != "" {
+			cfg.TLSConfig.CAFile = c.ConsulServers.CACertFile
+		}
+
+		if !strings.HasPrefix(c.ConsulServers.Hosts, "exec=") {
+			cfg.TLSConfig.Address = c.ConsulServers.Hosts
+		}
```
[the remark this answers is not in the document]
Ganeshrockz: Good catch. I had this change planned but somehow forgot to add it. Thanks for pointing out

config/config.go
```diff
@@ -0,0 +1,224 @@
...
+func (c *Config) ConsulServerConnMgrConfig(taskMeta awsutil.ECSTaskMeta) (discovery.Config, error) {
+	cfg := discovery.Config{
+		Addresses: c.ConsulServers.Hosts,
+		GRPCPort:  c.ConsulServers.GRPCPort,
+	}
+
+	if c.ConsulServers.EnableTLS {
+		tlsConfig := &tls.Config{}
+
+		caCert := os.Getenv(consulCACertPemEnvVar)
```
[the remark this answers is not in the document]
Ganeshrockz: Thanks for the detailed explanation. Makes sense. This commit https://github.com/hashicorp/consul-ecs/pull/143/commits/c1f0317e41a6849b34fa636417132a840387c099 should fix these concerns

config/config.go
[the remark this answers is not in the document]
pglass: Oh, we do need to specify the partition in the login details. The controller creates an auth method in the controller's assigned partition:

* We use `c.writeOptions()` here: https://github.com/hashicorp/consul-ecs/blob/91945d652eea4f78d2972b6c5beeff026fb40ce3/subcommand/acl-controller/command.go#L420
* The write options include the partition: https://github.com/hashicorp/consul-ecs/blob/91945d652eea4f78d2972b6c5beeff026fb40ce3/subcommand/acl-controller/command.go#L594-L599

subcommand/mesh-init/command.go
pglass: That's true. I think there are a few reasons for that:

1. Many consul-k8s components are short-lived (like jobs or init containers) so they use `State` once and don't need to watch for changes.
2. If gRPC is the only protocol (like in Consul Dataplane) then the gRPC connection is shared with consul-server-connection-manager and the gRPC connection automatically switches servers under the hood (no need to use State/Subscribe)
3. It's easy to use `State` to lazily construct a Consul client on-demand whenever a client is needed (consul-k8s does this in many places)

I added `Subscribe` to consul-server-connection-manager with long-running processes in mind, but it's fine to follow the consul-k8s style of using `State` everywhere (and then we could potentially remove `Subscribe` from consul-server-connection-manager it's not being used)

testutil/consul.go
```diff
@@ -56,3 +59,17 @@ func ConsulACLConfigFn(c *testutil.TestServerConfig) {
 	c.ACL.Tokens.InitialManagement = AdminToken
 	c.ACL.DefaultPolicy = "deny"
 }
+
+func GetHostAndPortFromAddress(address string) (string, int) {
```
pglass, line 63: minor: This function wouldn't handle IPv6 IPs correctly. We could use https://pkg.go.dev/net#SplitHostPort instead (although I think we generally assume IPv4 in ECS and don't test with IPV6. But it is [supported](https://aws.amazon.com/about-aws/whats-new/2020/11/amazon-ecs-supports-ipv6-in-awsvpc-networking-mode/) in ECS).

pglass (approved):
The one comment that is a "must fix" is setting the partition in the login data.

The rest of the comments are minor/non-blocking (feel free to address/ignore how you feel is best).

Otherwise, approved so you can merge when ready. This is great work!

config/resources/test_config_null_nested_fields.json
```diff
@@ -5,11 +5,23 @@
...
+  "consulServers": {
+    "hosts": "",
+    "https": null,
+    "httpPort": null,
+    "grpcPort": null,
+    "httpsCACertFile": null,
+    "caCertFile": null,
+    "tls": null,
+    "tlsServerName": null,
+    "skipServerWatch": null
```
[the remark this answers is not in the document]
Ganeshrockz: > I think consul-k8s assumes the same TLS settings for https and grpc+tls to the Consul servers?

We support configuring individual TLS CA certs for HTTP and gRPC in the current version of Consul ECS. The mesh task's terraform module exposes two variables to control that. So I guess we should continue maintaining both the settings going forward as well. What do you think?

That said, I do like the way you restructured these inputs so that users can go with a default setting for both the protocols. I will merge this PR in (since it has already become huge) and raise a separate one to address this restructuring.

pglass:
I suspect the CI tests are failing with Consul 1.13 because of gRPC TLS settings that were weird in 1.13.

You could try to fix this, or we can drop tests for Consul 1.13. We typically maintain and test with N-2 versions of Consul. So with Consul 1.16 GA that would be 1.16, 1.15, and 1.14.

Ganeshrockz:
> We typically maintain and test with N-2 versions of Consul.

If that's the case, I can merge this PR first and later on drop the tests only Consul 1.16 becomes GA. Since this isn't getting to main I think it should be fine.

PR #144, opened 2023-06-26
Ability to register services, gateway and proxy

Ganeshrockz:
- Added ability to register a service, sidecar proxy and gateways directly towards the Consul servers.
- Removed the `checks` nested field from the config schema because we don't support running health checks in the new architecture. Instead, we will sync the health (ECS health status) of ECS containers into Consul.
- In addition to the health sync containers, we also add a custom health check that indicates the `consul-dataplane` container's health status.
- As of this point the health status is set to critical when the services are registered. In the upcoming PRs, the consul ECS control plane will take care of syncing the health status periodically to Consul.
- Removed the execution of `consul connect envoy` command. This will be replaced by writing the dataplane startup configuration in the upcoming PRs
- Removed the previous version of the consul client that talked to the Consul agent present in the task.

## How I've tested this PR:

Unit tests. The E2E integration will still be broken even after this change.

## Checklist:
- [X] Tests added
- [ ] CHANGELOG entry added (Not needed for this change)


subcommand/mesh-init/command.go
```diff
@@ -380,31 +293,32 @@ func (c *Command) constructGatewayProxyRegistration(taskMeta awsutil.ECSTaskMeta
...
-	gwRegistration.Checks = []*api.AgentServiceCheck{
-		{
-			Name:                           fmt.Sprintf("%s listener", gwRegistration.Kind),
-			TCP:                            net.JoinHostPort(healthCheckAddr.Address, fmt.Sprint(healthCheckAddr.Port)),
-			Interval:                       "10s",
-			DeregisterCriticalServiceAfter: "10m",
-		},
+func getNodeMeta() map[string]string {
+	return map[string]string{
+		config.SyntheticNode: "true",
```
Ganeshrockz, line 322: consul-k8s follows a similar model for registering nodes. I guess we should add extra meta-data about the node to indicate that it comes from ECS. Currently the UI shows a kubernetes specific message if there are any synthetic nodes

pglass: Screenshot 2023-06-26 at 9 14 34 AM

Ah, it looks like it's this message? Did you notice anything else Kubernetes-specific?

subcommand/mesh-init/command.go
```diff
@@ -104,99 +92,58 @@ func (c *Command) realRun() error {
...
 		c.log.Info("wrote ACL token to shared volume", "token-file", tokenFile)
 	}
 
-	// Client config for the V2 client that talks directly to the
-	// server agent
-	consulClientCfg := c.config.ClientConfig()
-	consulClientCfg.Address = fmt.Sprintf("%s:%d", state.Address.IP.String(), c.config.ConsulServers.HTTPPort)
+	// Client config for the client that talks directly to the server agent
+	cfg := c.config.ClientConfig()
+	cfg.Address = fmt.Sprintf("%s:%d", state.Address.IP.String(), c.config.ConsulServers.HTTPPort)
```
pglass, line 100: minor: consider using https://pkg.go.dev/net#JoinHostPort (in case of an IPv6 address it will bracket the address as-needed)

pglass (approved):
👍 A few minor things

I think we can remove the `ecs-synthetic-node` metadata field until we know we need it.

config/types.go
```diff
@@ -33,13 +33,19 @@ const (
...
 	// Match Consul: https://github.com/hashicorp/consul/blob/68e79b8180ca89e8cfca291b40a30d943039bd49/agent/consul/authmethod/awsauth/aws.go#L16-L20
 	AuthMethodType         string = "aws-iam"
 	IAMServerIDHeaderName  string = "X-Consul-IAM-ServerID"
 	GetEntityMethodHeader  string = "X-Consul-IAM-GetEntity-Method"
 	GetEntityURLHeader     string = "X-Consul-IAM-GetEntity-URL"
 	GetEntityHeadersHeader string = "X-Consul-IAM-GetEntity-Headers"
 	GetEntityBodyHeader    string = "X-Consul-IAM-GetEntity-Body"
+
+	SyntheticNode    string = "synthetic-node"
+	ECSSyntheticNode string = "ecs-synthetic-node"
```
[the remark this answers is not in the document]
Ganeshrockz: My thought behind introducing this field is to help the UI show a ECS specific message similar to k8s. May be the banner now needs to show a generic image rather than a k8s specific one. I can remove it for now

config/schema.json
```diff
@@ -169,125 +169,6 @@
           "required": ["passing", "warning"],
           "additionalProperties": false
         },
-        "checks": {
```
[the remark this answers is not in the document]
Ganeshrockz: Do you know why `additionalProperties` was set to true only for this block?  Changing it to `false` for now and merging this PR. Added a backing test as well to make sure people don't change this in the future. Let me know if you think `additionalProperties` should be `true` for some reason for the service block. I can revert it in the next PR

PR #145, opened 2023-06-26
Added ability to generate dataplane configuration 

Ganeshrockz:
- With the newly proposed dataplane based architecture, we would want to make the  consul ECS control plane generate the configuration needed for dataplane to startup. This PR adds the necessary modules needed to generate a byte encoded JSON string. This will be written to a shared volume by the control plane which can then be consumed by the dataplane (This will be done in a later PR).
- The configuration JSON looks like

```json
{
   "consul": {
          "addresses": "", // Same as `consulServers.addresses`
          "grpcPort": 8503, // Same as `consulServers.grpcPort`
          "serverWatchDisabled": false, // Same as `consulServers.skipServerWatch`
          "tls": {
                "disabled" : false, // Defaults to false.
                "caCertsPath": "", // Path of the CA cert's PEM for securing internal grpc calls
                "tlsServerName": "" // Same as `consulServers.tlsServerName
           },
         "credentials": {
               "type": "static", //Defaults to static,
               "static": {
                   "token": "" // ACL token
               }
         }
    },
    "service": {
         "nodeName": "", // Same as cluster ARN
         "serviceID": "", // ID of the sidecar proxy
         "namespace": "",
         "partition": ""
     },
     "xdsServer": {
          "bindAddress": "127.0.0.1",
          "bindPort": 20000,
      }
}
``` 

## How I've tested this PR:

Unit tests

## Checklist:
- [X] Tests added
- [ ] CHANGELOG entry added (Not needed)


internal/dataplane/dataplane_tpl.go
```diff
@@ -0,0 +1,226 @@
...
+		"adminBindAddress": "` + e.adminBindAddress + `",
+		"adminBindPort": ` + strconv.FormatInt(int64(e.adminBindPort), 10) + `
+	}`
+}
+
+type telemetryConfig struct {
+	useCentralConfig bool
+}
+
+func (t telemetryConfig) generateJSON() string {
```
Ganeshrockz, line 222: Should we add prometheus related stuff to this JSON? I see that those settings are overridable by ENV vars and so decided to leave it like this with the value of `useCentralConfig` as false

internal/dataplane/dataplane_tpl.go
```diff
@@ -0,0 +1,226 @@
...
+type dataplaneConfig struct {
+	consul    consulConfig
+	service   serviceConfig
+	logging   loggingConfig
+	xdsServer xdsServerConfig
+	envoy     envoyConfig
+	telemetry telemetryConfig
+}
+
+type dataplaneConfigTmplArgs struct {
```
pglass, line 19: Instead of using templating, do you think we can rely only on `encoding/json`, possibly with custom `MarshalJSON` method implementations where necessary (for including/excluding fields)?

internal/dataplane/dataplane_tpl.go
pglass: I actually think we should not set default values here that are different than consul-dataplane 🤔 (except where it really makes sense to do otherwise).

So I'd opt to remove the telemetryConfig here, and consul-dataplane will default to `true` for this.

(I know I wrote in the RFC that this would default to false, but that was before consul-dataplane was actually implemented, and we ended up with `-use-central-config=true` as the default in Consul Dataplane.)

internal/dataplane/dataplane_tpl.go
```diff
@@ -0,0 +1,226 @@
...
+}
+
+func (x xdsServerConfig) generateJSON() string {
+	return `{
+		"bindAddress": "` + x.address + `",
+		"bindPort": ` + strconv.FormatInt(int64(x.port), 10) + `
+	}`
+}
+
+type envoyConfig struct {
```
pglass, line 206: We can probably remove this `envoyConfig` struct, because consul-dataplane defaults to `127.0.0.1:19000`.

Users can change this with environment variables, if needed (`DP_ENVOY_ADMIN_BIND_ADDRESS`, `DP_ENVOY_ADMIN_BIND_PORT`).

internal/dataplane/dataplane_tpl.go
```diff
@@ -0,0 +1,226 @@
...
+
+type loggingConfig struct {
+	json  bool
+	level string
+}
+
+func (l loggingConfig) generateJSON() string {
+	return `{
+		"logLevel": "` + l.level + `",
+		"logJSON": ` + strconv.FormatBool(l.json) + `
```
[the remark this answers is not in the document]
pglass: Or, if environment variables take precedence over the config file field, then it's fine to keep this.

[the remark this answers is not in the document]
Ganeshrockz: Yeah. Environment variables take precedence over file fields. `logJSON` defaults to false in dataplane. I'll remove the field

internal/dataplane/dataplane_tpl.go
Ganeshrockz: Thanks for pointing it out. Using `json.Marshal` makes it much more easier to generate the JSON. Removed the old behaviour based on templating

pglass (approved):

PR #147, opened 2023-07-03
Support to gracefully shutdown control-plane

Ganeshrockz:
- Adds a monitor to keep track of the dataplane container's status. This is needed to gracefully shutdown the control plane after the dataplane is shutdown. The monitor only wakes up after SIGTERM and polls ECS to get the status of the dataplane container. Once it has confirmed that the dataplane container has stopped, it signals control plane to begin it's graceful shutdown flow.
- Adds support to deregister proxies and services during shutdown
- Adds support for consul logout
- Adds support to cleanup all the config files from the `bootstrapDirectory`

## How I've tested this PR:

Unit tests

## Checklist:
- [X] Tests added
- [ ] CHANGELOG entry added(Not needed)

### DIFF
--- a/subcommand/mesh-init/command.go
+++ b/subcommand/mesh-init/command.go
@@ -259,12 +259,6 @@ func (c *Command) realRun() error {
 				}
 			}
 
-			// Cleanup bootstrap directory
-			err = os.RemoveAll(c.config.BootstrapDir)
-			if err != nil {
-				c.log.Error("error cleaning up bootstrap directory %s", err.Error())
-				result = multierror.Append(result, err)
-			}
 			return result
 		}
 	}
--- a/subcommand/mesh-init/command.go
+++ b/subcommand/mesh-init/command.go
@@ -496,13 +496,21 @@ func (c *Command) generateAndWriteDataplaneConfig(proxyRegistration *api.Catalog
 }
 
 func (c *Command) deregisterServiceAndProxy(consulClient *api.Client, clusterARN string, serviceRegistration, proxyRegistration *api.CatalogRegistration) error {
+	var result error
 	if serviceRegistration != nil {
-		deregisterConsulService(consulClient, serviceRegistration, clusterARN)
+		err := deregisterConsulService(consulClient, serviceRegistration, clusterARN)
+		if err != nil {
+			result = multierror.Append(result, err)
+		}
 	}
 
 	// Proxy deregistration
-	deregisterConsulService(consulClient, proxyRegistration, clusterARN)
-	return nil
+	err := deregisterConsulService(consulClient, proxyRegistration, clusterARN)
+	if err != nil {
+		result = multierror.Append(result, err)
+	}
+
+	return result
 }
 
 func deregisterConsulService(client *api.Client, reg *api.CatalogRegistration, node string) error {
@@ -514,11 +522,7 @@ func deregisterConsulService(client *api.Client, reg *api.CatalogRegistration, n
 	}
 
 	_, err := client.Catalog().Deregister(deregInput, nil)
-	if err != nil {
-		return err
-	}
-
-	return nil
+	return err
 }
 
 func getNodeMeta() map[string]string {
--- a/subcommand/mesh-init/command.go
+++ b/subcommand/mesh-init/command.go
@@ -27,6 +27,7 @@ import (
 	"github.com/hashicorp/consul-server-connection-manager/discovery"
 	"github.com/hashicorp/consul/api"
 	"github.com/hashicorp/go-hclog"
+	"github.com/hashicorp/go-multierror"
 	"github.com/mitchellh/cli"
 )
 
@@ -42,6 +43,8 @@ type Command struct {
 
 	isHealthy atomic.Bool
 	checks    map[string]*api.HealthCheck
+
+	dataplaneMonitor *dataplaneMonitor
 
 	// Following fields are only needed for unit tests
 
@@ -89,6 +92,7 @@ func (c *Command) Run(args []string) int {
 	c.config = config
 
 	c.log = logging.FromConfig(c.config).Logger()
+	c.dataplaneMonitor = newDataplaneMonitor(c.ctx, c.log)
 
 	err = c.realRun()
 	if err != nil {
@@ -104,6 +108,8 @@ func (c *Command) realRun() error {
 
 	// Register and start health check handler.
 	go c.startHealthCheckServer()
+
+	go c.dataplaneMonitor.run()
 
 	taskMeta, err := awsutil.ECSTaskMetadata()
 	if err != nil {
@@ -223,7 +229,6 @@ func (c *Command) realRun() error {
 		<-c.proceedChan
 	}
 
-	// TODO: Handle graceful shutdown
 	for {
 		select {
 		case <-time.After(syncChecksInterval):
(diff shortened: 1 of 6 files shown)


subcommand/mesh-init/command.go
```diff
@@ -467,6 +495,32 @@ func (c *Command) generateAndWriteDataplaneConfig(proxyRegistration *api.Catalog
...
 }
 
+func (c *Command) deregisterServiceAndProxy(consulClient *api.Client, clusterARN string, serviceRegistration, proxyRegistration *api.CatalogRegistration) error {
+	if serviceRegistration != nil {
+		deregisterConsulService(consulClient, serviceRegistration, clusterARN)
+	}
+
+	// Proxy deregistration
+	deregisterConsulService(consulClient, proxyRegistration, clusterARN)
+	return nil
```
johnlanda, line 505: it looks like we never return the error here. Should this be the following?

```suggestion
	if serviceRegistration != nil {
		return deregisterConsulService(consulClient, serviceRegistration, clusterARN)
	}

	// Proxy deregistration
	return deregisterConsulService(consulClient, proxyRegistration, clusterARN)
```

subcommand/mesh-init/command.go
```diff
@@ -467,6 +495,32 @@ func (c *Command) generateAndWriteDataplaneConfig(proxyRegistration *api.Catalog
...
+		Namespace: reg.Service.Namespace,
+		Partition: reg.Service.Partition,
+	}
+
+	_, err := client.Catalog().Deregister(deregInput, nil)
+	if err != nil {
+		return err
+	}
+
+	return nil
```
johnlanda, line 521:
```suggestion
	_, err := client.Catalog().Deregister(deregInput, nil)
        return err
```

subcommand/mesh-init/command_test.go
```diff
@@ -972,6 +1009,24 @@ func constructContainerResponse(name, health string) awsutil.ECSTaskMetaContaine
...
+// task meta response
+func stopDataplaneContainer(taskMetadataResp *awsutil.ECSTaskMeta) {
+	index := -1
+	for i, c := range taskMetadataResp.Containers {
+		if isDataplaneContainer(c) {
+			index = i
+			break
+		}
+	}
+	if index == -1 {
```
johnlanda, line 1022: super minor nit: `if index >= -1` captures all possible invalid values for the index. I know it should never be anything other less than -1 but just a convention I am used to as it is a more constrained condition.

subcommand/mesh-init/dataplane-monitor.go
```diff
@@ -0,0 +1,99 @@
...
+	d.log.Info("waiting for dataplane container to stop")
+	for {
+		select {
+		case <-d.ctx.Done():
+			return
+		case <-time.After(1 * time.Second):
+			taskMeta, err := awsutil.ECSTaskMetadata()
+			if err != nil {
+				d.log.Error("fetching task metadata", "err", err.Error())
+				break
```
johnlanda, line 55: What happens to anything waiting on the `doneCh` if we exit here due to an error getting the `ECSTaskMetadata` or if we return due to `d.ctx.Done()`?

subcommand/mesh-init/command_test.go
Ganeshrockz: Should this be index<=-1?

subcommand/mesh-init/dataplane-monitor.go
Ganeshrockz: I have added a `defer close(doneCh)` at the start of this function. Wouldn't that take of care of the cases you mentioned?

subcommand/mesh-init/command_test.go
johnlanda: Oops sorry typo. Yes `index <= -1`

subcommand/mesh-init/dataplane-monitor.go
johnlanda: Yes I think that covers it. Missed that the first time around.

subcommand/mesh-init/command_test.go
```diff
@@ -972,6 +1009,24 @@ func constructContainerResponse(name, health string) awsutil.ECSTaskMetaContaine
...
 }
 
+// stopDataplaneContainer marks the dataplane container's status as STOPPED in the
+// task meta response
+func stopDataplaneContainer(taskMetadataResp *awsutil.ECSTaskMeta) {
+	index := -1
+	for i, c := range taskMetadataResp.Containers {
+		if isDataplaneContainer(c) {
+			index = i
+			break
```
pglass, line 1019: nit: simplification - do the work in the array and return (so we don't need to track a found index)

```suggestion
			taskMetadataResp.Containers[i].DesiredStatus = ecs.DesiredStatusStopped
			taskMetadataResp.Containers[i].KnownStatus = ecs.DesiredStatusStopped
			return
```

And if `c` is not a copy, then this could just be `c.DesiredStatus = ...` instead of reindexing the array.

Another option is [`slices.IndexFunc`](https://pkg.go.dev/golang.org/x/exp/slices#IndexFunc) (note: experimental package but fine for tests) where I think this loop could be replaced by `index = slices.IndexFunc(taskMetadataResp.Containers, isDataplaneContainer)`

subcommand/mesh-init/dataplane-monitor.go
```diff
@@ -0,0 +1,99 @@
...
+	log hclog.Logger
+
+	doneCh chan struct{}
+}
+
+func newDataplaneMonitor(ctx context.Context, logger hclog.Logger) *dataplaneMonitor {
+	return &dataplaneMonitor{
+		ctx:    ctx,
+		log:    logger,
+		doneCh: make(chan struct{}, 1),
```
pglass, line 27: nit: I think this channel can be unbuffered

```suggestion
		doneCh: make(chan struct{}),
```

subcommand/mesh-init/dataplane-monitor.go
```diff
@@ -0,0 +1,99 @@
...
+}
+
+// Run will wake up when SIGTERM is received. Then, it polls task metadata
+// until the dataplane container stops. Use the Done() channel to wait
+// until it has finished.
+func (d *dataplaneMonitor) run() {
+	defer close(d.doneCh)
+
+	if !d.waitForSIGTERM() {
+		d.doneCh <- struct{}{}
```
pglass, line 42: minor: It should be enough to close the channel to indicate done-ness, so we probably don't actually need to send anything on the channel. I think we can just return and the `defer close(d.doneCh)` at the top will take care of it for us.

What do you think?

```suggestion
```

subcommand/mesh-init/dataplane-monitor.go
```diff
@@ -0,0 +1,99 @@
...
+		case <-time.After(1 * time.Second):
+			taskMeta, err := awsutil.ECSTaskMetadata()
+			if err != nil {
+				d.log.Error("fetching task metadata", "err", err.Error())
+				break
+			}
+
+			if hasDataplaneContainerStopped(taskMeta) {
+				d.log.Info("dataplane container has stopped, terminating control plane")
+				d.doneCh <- struct{}{}
```
pglass, line 60: Same here: I think closing the channel (in the defer when we return) is sufficient and we don't need to send on the channel.

subcommand/mesh-init/command.go
```diff
@@ -235,9 +240,32 @@ func (c *Command) realRun() error {
...
+					result = multierror.Append(result, err)
+				}
+			}
+
+			// Cleanup bootstrap directory
+			err = os.RemoveAll(c.config.BootstrapDir)
+			if err != nil {
+				c.log.Error("error cleaning up bootstrap directory %s", err.Error())
+				result = multierror.Append(result, err)
+			}
```
pglass, line 267: Why remove the bootstrap directory? I feel like we can skip this step. The task is stopping and the data in the volume will be removed by ECS anyway, so we shouldn't need to do it ([ref1](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/fargate-task-storage.html), [ref2](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/bind-mounts.html)). What do you think?

subcommand/mesh-init/dataplane-monitor.go
```diff
@@ -0,0 +1,99 @@
...
+}
+
+func isDataplaneContainer(container awsutil.ECSTaskMetaContainer) bool {
+	return container.Name == config.ConsulDataplaneContainerName
+}
+
+func isStopped(container awsutil.ECSTaskMetaContainer) bool {
+	return container.DesiredStatus == ecs.DesiredStatusStopped &&
+		container.KnownStatus == ecs.DesiredStatusStopped
+}
```
pglass, line 99: suggestion: Add methods for these types of things to the `awsutil.ECSTaskMeta` type. Maybe `ECSTaskMeta.FindContainer` and `ECSTaskMetaContainer.IsStopped` or etc.

pglass (approved):
👍

PR #148, opened 2023-07-07
Deprecate mesh-init in favour of control-plane

Ganeshrockz:
- Just renames the `mesh-init` package names to `control-plane`
- The mesh-init command is also replaced by the `control-plane` command

## Checklist:
- [ ] Tests added - Not needed since this is just a rename
- [ ] CHANGELOG entry added

### DIFF
--- a/config/schema.json
+++ b/config/schema.json
@@ -11,7 +11,7 @@
       "enum": ["TRACE", "DEBUG", "INFO", "WARN", "ERROR", null]
     },
     "bootstrapDir": {
-      "description": "The directory at which to mount the shared volume where dataplane configuration is written by `consul-ecs control-plane`.",
+      "description": "The directory at which to mount the shared volume where Consul dataplane configuration is written by `consul-ecs control-plane`.",
       "type": "string",
       "minLength": 1
     },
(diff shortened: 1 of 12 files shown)


config/schema.json
```diff
@@ -6,12 +6,12 @@
...
   "properties": {
     "logLevel": {
-      "description": "Sets the log level for the `consul-ecs mesh-init` and `consul-ecs health-sync` commands. Defaults to `INFO`.",
+      "description": "Sets the log level for the `consul-ecs control-plane` and `consul-ecs health-sync` commands. Defaults to `INFO`.",
       "type": ["string", "null"],
       "enum": ["TRACE", "DEBUG", "INFO", "WARN", "ERROR", null]
     },
     "bootstrapDir": {
-      "description": "The directory at which to mount the shared volume where Envoy bootstrap configuration is written by `consul-ecs mesh-init`.",
+      "description": "The directory at which to mount the shared volume where dataplane configuration is written by `consul-ecs control-plane`.",
```
pglass, line 14: minor: I think "dataplane" on its own can be ambiguous. Better to call it "consul-dataplane" or "Consul Dataplane" here.

```suggestion
      "description": "The directory at which to mount the shared volume where consul-dataplane configuration is written by `consul-ecs control-plane`.",
```

pglass (approved):

PR #149, opened 2023-07-10
Server watch for agentless

Ganeshrockz:
- Implemented changes to enable server watch in control plane
- Makes use of the `Subscribe()` function in connection manager to listen to changes in the servers. If servers get changed, we make sure to write the new ACL token in the same shared volume and reconfigure the consul client accordingly

## How I've tested this PR:

Unit tests

## Checklist:
- [X] Tests added
- [ ] CHANGELOG entry added


Ganeshrockz:
I would need some feedback from reviewers on how to test this. From what I see, there are way options to proceed

1. Shutdown the consul server within the testcase and restart it. Our current test setup runs tests against a single consul server replica. The main issue here is that the moment the server restarts, it occupies a different set of GRPC and HTTP ports compared to the initial one which were passed as inputs to the `CONSUL_ECS_CONFIG` environment var. Currently the test server picks up the free ports available to it during the initialization time which makes it impossible to test the E2E with server watch enabled.

2. Mock the server rediscovery behaviour by sending a signal to the `watcherCh` which is being listened to by the control plane in the reconciliation loop. We will not shutdown the server in this option, but will send the same `discovery.State` data for the original server. The `watcherCh` variable will not subscribe to the connection manager in case of unit tests.

I have chosen the second option for this PR because option 1 seemed impossible to implement. Let me know if there is an easier way to approach the E2E test flow in this case.

subcommand/control-plane/command_test.go
```diff
@@ -498,6 +519,15 @@ func TestRun(t *testing.T) {
 			// Verify with retries that the checks have reached the expected state
 			assertHealthChecks(t, consulClient, expectedServiceChecks, expectedProxyCheck)
 
+			if !c.skipServerWatch {
+				addr, err := discovery.MakeAddr(serverHost, serverGRPCPort)
+				require.NoError(t, err)
+
+				watcherCh <- discovery.State{
+					Address: addr,
+				}
```
pglass, line 528: Responding to your other comment here, so that conversation is threaded

> I would need some feedback from reviewers on how to test this. From what I see, there are way options to proceed
> 
> 1. Shutdown the consul server within the testcase and restart it. Our current test setup runs tests against a single consul server replica. The main issue here is that the moment the server restarts, it occupies a different set of GRPC and HTTP ports compared to the initial one which were passed as inputs to the CONSUL_ECS_CONFIG environment var. Currently the test server picks up the free ports available to it during the initialization time which makes it impossible to test the E2E with server watch enabled.
> 
> 2. Mock the server rediscovery behaviour by sending a signal to the watcherCh which is being listened to by the control plane in the reconciliation loop. We will not shutdown the server in this option, but will send the same discovery.State data for the original server. The watcherCh variable will not subscribe to the connection manager in case of unit tests.
> 
> I have chosen the second option for this PR because option 1 seemed impossible to implement. Let me know if there is an easier way to approach the E2E test flow in this case.

It sounds like you want to test the flow when the connection switches Consul servers?

I think (2) is fine, and probably easier. It's also generally fine to mock, because we don't necessarily have to test the functionality of consul-server-connection-manager here (although that would be nice for testing completeness).

pglass: My only comment is to consider introducing an interface type so that we can mock the whole discovery.Watcher type, rather than just the channel. Similar to what is done in consul-k8s:

* Interface: https://github.com/hashicorp/consul-k8s/blob/fd201c57caa59875492914392b6e45a368e3d749/control-plane/consul/consul.go#L16-L22 
* Generated mock: https://github.com/hashicorp/consul-k8s/blob/fd201c57caa59875492914392b6e45a368e3d749/control-plane/consul/mock_ServerConnectionManager.go#L10-L13

We could also introduce that interface and generated mock into consul-server-connection-manager, to generally make this easier to test in general.

What do you think? This is non-blocking - just thinking about potential improvements.

---

also, for more context: What I did in consul-server-connection-manager is [start a multi-server Consul cluster](https://github.com/hashicorp/consul-server-connection-manager/blob/bda59462159ced662b7e4018e0ac2b11aa16c7f1/discovery/watcher_test.go#L333-L362). Then I could [stop a server](https://github.com/hashicorp/consul-server-connection-manager/blob/bda59462159ced662b7e4018e0ac2b11aa16c7f1/discovery/watcher_test.go#L328) and check that the client switches to one of the other running servers. But, this required mocking some internals that aren't exported, because consul-server-connection-manager assumes the same gRPC port for all servers. So we could make fairly small changes consul-server-connection-manager to support better testing with "real" Consul test servers, rather than mocking - as a potential alternative / complement to above.

subcommand/control-plane/command.go
```diff
@@ -233,6 +216,14 @@ func (c *Command) realRun() error {
 		select {
 		case <-time.After(syncChecksInterval):
 			currentHealthStatuses = c.syncChecks(consulClient, currentHealthStatuses, serviceName, clusterARN, healthSyncContainers)
+		case watcherState := <-c.watcherCh:
+			c.log.Info("Consul server change detected. Reconfiguring consul client to connect to the newly discovered server(s)")
```
pglass, line 220: We could include the server address here for clarity:

```suggestion
			c.log.Info("Switching to Consul server", "address", watcherState.Addr.String())
```

subcommand/control-plane/command_test.go
Ganeshrockz: The mocked connection manager approach looks a lot cleaner. I've gone ahead and implemented the same. Thanks for the pointers.

pglass (approved):

PR #150, opened 2023-07-10
Controller changes  to support new ECS architecture - Part 1

Ganeshrockz:
- Exposes the `controller` field in the config JSON. With agentless, the `ecs-controller` will receive the entire config JSON instead of individual flags to facilitate server discovery and other stuff. The json looks like
```json
"controller": {
    "iamRolePath": "", // Defaults to `/consul-ecs/`
    "partitionsEnabled": false,
    "partition": ""
}
```
- Removed `client` specific policy creation, role creation, auth method creation from the existing ACL controller.
- Added ability to register clusterARN as the synthetic node for ECS in the catalog
- Added ability to discover servers in the controller. Server watch will be added in the upcoming PRs.
- Configured the client with the `CONSUL_HTTP_TOKEN` which is the bootstrap token we expect to have `acl:write`, `operator:write` and `node:write` permissions.
- Removed flags from the controller command

## How I've tested this PR:

Unit tests

## Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added

### DIFF
--- a/subcommand/acl-controller/command.go
+++ b/subcommand/acl-controller/command.go
@@ -103,12 +103,12 @@ func (c *Command) run() error {
 
 	serverConnMgrCfg, err := c.config.ConsulServerConnMgrConfig(ecsMeta)
 	if err != nil {
-		return fmt.Errorf("constructing server connection manager config: %s", err)
+		return fmt.Errorf("constructing server connection manager config: %w", err)
 	}
 
 	watcher, err := discovery.NewWatcher(c.ctx, serverConnMgrCfg, c.log)
 	if err != nil {
-		return fmt.Errorf("unable to create consul server watcher: %s", err)
+		return fmt.Errorf("unable to create consul server watcher: %w", err)
 	}
 
 	go watcher.Run()
@@ -116,7 +116,7 @@ func (c *Command) run() error {
 
 	state, err := watcher.State()
 	if err != nil {
-		return fmt.Errorf("unable to fetch consul server watcher state: %s", err)
+		return fmt.Errorf("unable to fetch consul server watcher state: %w", err)
 	}
 
 	// Client config for the client that talks directly to the server agent
@@ -130,7 +130,7 @@ func (c *Command) run() error {
 
 	consulClient, err := api.NewClient(cfg)
 	if err != nil {
-		return fmt.Errorf("constructing consul client from config: %s", err)
+		return fmt.Errorf("constructing Consul API client from config: %w", err)
 	}
 
 	if err := c.upsertConsulResources(consulClient, ecsMeta, clusterArn); err != nil {
@@ -229,7 +229,7 @@ func (c *Command) upsertConsulResources(consulClient *api.Client, ecsMeta awsuti
 			return err
 		}
 	} else if c.config.Controller.Partition != "" {
-		return fmt.Errorf("`config.controller.partition` provided without setting `config.controller.partitionEnabled` as true")
+		return fmt.Errorf("`config.controller.partition` provided without setting `config.controller.partitionEnabled = true`")
 	}
 
 	if err := c.upsertAuthMethod(consulClient, serviceAuthMethod); err != nil {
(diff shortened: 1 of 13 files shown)


subcommand/acl-controller/command_test.go
```diff
@@ -232,69 +259,25 @@ func checkConsulResources(t *testing.T, consulClient *api.Client, expPolicyRules
...
 	}
 
-	require.Contains(t, policyNames, "consul-ecs-client-policy")
 	if partitionsEnabled {
 		// We test with a non-default partition which lacks the global-management policy.
 		// The anonymous token policy is only created in the default partition, since that
-		// is where the anonymous token lives, so we expect only the client policy.
-		require.Len(t, policies, 1)
+		// is where the anonymous token lives, so we expect no policies to be present
+		require.Len(t, policies, 0)
```
pglass, line 266: minor/unrelated maintenance thing that is not your fault: We are adding a new built-in policy to Consul in the near future, and it is likely to go into 1.16, 1.15, and 1.14 patch releases. That means these assertions will likely break because there will be a different number of built-in policies that this expects.

We probably want to remove these `require.Len` assertions and only check whether the anonymous-token-policy was created or not. (I should have written it this way originally 🙃 ):

* If partitionsEnabled:
  * `policies` does not contain `"anonymous-token-policy"` (i.e. `require.NotContains(...)`)
* Else:
  * `policies` does contain `"anonymous-token-policy"`
 
Or we could just use the [Read Policy by Name API](https://developer.hashicorp.com/consul/api-docs/acl/policies#read-a-policy-by-name) to directly check for the anonymous-token-policy.

subcommand/acl-controller/command_test.go
```diff
@@ -223,7 +233,24 @@ func checkConsulResources(t *testing.T, consulClient *api.Client, expPolicyRules
 		require.Equal(t, partitions[1].Name, testPartitionName)
 	}
 
-	// Check the client policy is created.
+	// Check if the node got registered properly
+	partitionToCheck := ""
+	if partitionsEnabled {
+		partitionToCheck = testPartitionName
+	}
+	nodes, _, err := consulClient.Catalog().Nodes(&api.QueryOptions{Partition: partitionToCheck})
```
pglass, line 241: minor/nit: You could use an [API filter here](https://developer.hashicorp.com/consul/api-docs/catalog#filtering) to filter by Node name:

```suggestion
	nodes, _, err := consulClient.Catalog().Nodes(&api.QueryOptions{
		Partition: partitionToCheck,
		Filter:    fmt.Sprintf("Node == %q", clusterARN),
	})
```

And then you can just check `require.Len(t, nodes, 1)` instead of the following loop (and if you want to be extra sure, also `require.Equal(t, nodes[0].Node, clusterARN)`)

subcommand/acl-controller/command.go
```diff
@@ -213,44 +212,40 @@ func (c *Command) upsertConsulResources(consulClient *api.Client, ecsMeta awsuti
...
-	}
 	if err := c.upsertBindingRule(consulClient, serviceBindingRule); err != nil {
 		return err
 	}
 	if err := c.upsertAnonymousTokenPolicy(consulClient, agentConfig); err != nil {
 		return err
 	}
+
+	if err := c.registerNode(consulClient, ecsMeta, clusterARN); err != nil {
+		return fmt.Errorf("registering ecs cluster as a node in Consul: %s", err)
```
pglass, line 246: nit: Use `%w` for error wrapping:

```suggestion
		return fmt.Errorf("registering ECS cluster as a node in Consul: %w", err)
```

subcommand/acl-controller/command.go
```diff
@@ -102,27 +101,47 @@ func (c *Command) run() error {
...
+
+	token := os.Getenv(bootstrapTokenEnvVar)
+	if token != "" {
+		cfg.Token = token
 	}
 
 	consulClient, err := api.NewClient(cfg)
 	if err != nil {
-		return err
+		return fmt.Errorf("constructing consul client from config: %s", err)
```
pglass, line 133: nit: Use `%w` for error wrapping. Also clarity suggestion ("consul client" vs "consul api client")

```suggestion
		return fmt.Errorf("constructing Consul API client from config: %w", err)
```

pglass (approved):
Approved, just a few minor things

subcommand/acl-controller/command_test.go
Ganeshrockz: Fixed this

PR #152, opened 2023-07-11
Deprecate health-sync command

Ganeshrockz:
- Deprecated the `health-sync` command since the health sync logic moves into the control plane in the new ECS architecture.
- Fixed ACL based unit tests for consul-ecs control plane

## How I've tested this PR:

Units tests

## Checklist:
- [X] Tests added
- [ ] CHANGELOG entry added - Not needed


subcommand/control-plane/command_test.go
```diff
@@ -1026,7 +1081,7 @@ func getExpectedDataplaneCfgJSON() string {
 	"consul": {
 	  "addresses": "127.0.0.1",
 	  "grpcPort": %d,
-	  "serverWatchDisabled": false
+	  "serverWatchDisabled": false%s
```
cthain, line 1084: Is this a typo?

```suggestion
	  "serverWatchDisabled": false
```

commands.go
```diff
@@ -33,9 +32,6 @@ func init() {
 		"acl-controller": func() (cli.Command, error) {
 			return &cmdController.Command{UI: ui}, nil
 		},
-		"health-sync": func() (cli.Command, error) {
-			return &cmdHealthSync.Command{UI: ui}, nil
-		},
```
cthain, line 38: nit: This isn't deprecating the `health-sync` command, it is removing it completely. I think this is a breaking change so we should ensure that the changelog gets updated accordingly.

cthain (commented):
Looks good :+1: I left a couple of minor comments.

commands.go
pglass: fwiw, this is merging into a non-main / feature branch, so we can either add the changelog entry now or when this feature branch goes into main. I'm fine either way.

But agreed! There are many breaking changes for agentless that need to be highlighted in the changelog, and it would probably be good to start adding those to the changelog on the feature branch to keep track of them.

subcommand/control-plane/command_test.go
Ganeshrockz: No. This was added to accommodate the `credentials` field when consul login is enabled. ([ref](https://github.com/hashicorp/consul-ecs/blob/1903328205d099c42d48e940fbeee49430db8f26/subcommand/control-plane/command_test.go#L941))

commands.go
Ganeshrockz: The whole of agentless is a breaking change to Consul ECS. I am planning to accumulate all the changes from all the PRs raised against the feature branch and add them in the final PR that gets merged to the main branch.

subcommand/control-plane/command_test.go
cthain: :+1: Got it, thanks.

Ganeshrockz:
@cthain Can you approve the PR if there are no more concerns?

cthain (approved):

subcommand/control-plane/command_test.go
Ganeshrockz, line 1: @pglass I had to remove the mocks added for server connection manager to test E2E token generation flows in this PR for the following reasons

1. Mocking server connection manager would mean that we should mock the ACL token that gets returned as well. If we send back the management token, then testing Consul logout flow becomes impossible because `client.ACL().TokenReadSelf(nil)` does not return an error for a client configured with the management token.
2. Mocking the token also reduces test coverage for `control-plane` because the flows related to constructing discovery credentials for connection manager become irrelevant because we always rely on the mocked server connection manager returning the management token.

For now, I am triggering a server address change by sending data to the `watcherCh` channel from the tests. This will enable us to test server watch and consul login/logout flows seamlessly. As a future improvement, we could make some small changes to connection manager to make the E2E testable in a seamless manner.


I am merging this PR in the current state. I am fine with reiterating on this in the upcoming PRs if you feel otherwise.

pglass: This is fine, and I actually prefer not mocking whenever possible. I thought a re-usable mock of the Watcher type might be nicer/easier as you start to test the controller as well, but it sounds like not. Thanks for looking into this!

PR #153, opened 2023-07-12
Controller changes to support new architecture - Part 2

Ganeshrockz:
- Refactored TaskStateLister to  build consul clients on demand
- Enable the controller to reconcile services during its reconciliation loop
- Renamed `acl-controller` to `controller` since it holds additional responsibilities like node registration and service deregistration in the new architecture

## How I've tested this PR:

Unit tests

## Checklist:
- [X] Tests added
- [ ] CHANGELOG entry added


config/types.go
```diff
@@ -36,6 +36,10 @@ const (
 	// Name of the dataplane's container
 	ConsulDataplaneContainerName = "consul-dataplane"
 
+	// Indicates that a service was registered by a control plane container running in
+	// a mesh task managed by Consul dataplane
+	DataplaneBasedMeshTaskTag = "dataplane-based-mesh-task"
```
Ganeshrockz, line 41: Chose such a name to avoid exposing the `agentless` to the public. Open for feedback on this.

controller/resource.go
```diff
@@ -343,6 +352,66 @@ func (s TaskStateLister) createNamespaces(resources []Resource) error {
...
 }
 
+// ReconcileServices ensures services are deregistered for missing ECS tasks with the
+// following steps
+//
+//   - Query the list of services belonging to the clusterARN node from Consul
+//   - Filter out the services that are registered by a Consul ECS control plane container
+//     running in a mesh task without a Consul client container.
+//   - Deregister those services for which the backing task resource is not available in ECS.
+func (s TaskStateLister) ReconcileServices(resources []Resource) error {
```
pglass, line 362: Can we have `TaskStateLister.List()` discover service instances instead of adding `ReconcileServices` to the interface?

That should fit into the existing interfaces:

* `TaskStatelist.List()` produces a list of `Resource`s (which are `TaskState`s) to reconcile.
   * A TaskState represents an ECS task and should have all the fields necessary to know the ACL tokens associated with the task, the service ids in Consul associated with the task, and whether those need to be cleaned up (i.e. whether the task is still running)
* Then each `Resource` can be independently reconciled - to delete tokens and service instances

I think we have `ReconcileNamespaces` as an exception because there is a `many:few` mapping of Resources to Namespaces, so it would be inefficient to reconcile namespaces as part of each TaskState.Reconcile.

But in this case, each task directly maps to a couple of service instances and those can be independently reconciled, and I don't think we need to deviate from the existing interfaces.

What do you think?

subcommand/acl-controller/command.go
```diff
@@ -137,30 +133,64 @@ func (c *Command) run() error {
...
+			c.log.Info("Switching to Consul server", "address", watcherState.Address.String())
+			consulClient, err = c.setupConsulAPIClient(watcherState)
+			if err != nil {
+				c.log.Error("error re-configuring consul client %s", err.Error())
+			} else {
+				ctrl = c.setupController(ecsClient, consulClient, clusterArn)
+			}
+		case <-c.ctx.Done():
+			return nil
+		}
```
pglass, line 156: I'm wondering if we can keep the `Controller.Run` method. I feel like doing so reduces the size of this large-ish `run()` function in the subcommand, and moves this loop functionality into the Controller where the loop logic is more easily testable in isolation.

This is probably a case where it would be nicer to have TaskStateLister call `watcher.State()` (compared to using Subscribe) in order to simply rebuild the api client on-demand. Then the Controller type itself doesn't have to be aware of the Watcher at all. Only the ResourceLister implementations do.

Potentially, we could change TaskStateLister to look like:

```go
// TaskStateLister is an implementation of ResourceLister.
type TaskStateLister struct {
	MakeClientFn func() (*api.Client, error)
	...
}
```

And then when we construct the TaskStateLister, we can do:

```go
	taskStateLister := &controller.TaskStateLister{
		ECSClient:  ecsClient,
		MakeClientFn: c.setupConsulAPIClient,
		ClusterARN: clusterARN,
		Partition:  c.config.Controller.Partition,
		Log:        c.log,
	}
```

And that would need a small change to `setupConsulAPIClient()`:

```go
func (c *Command) setupConsulAPIClient() (*api.Client, error) {
	state, err := c.watcher.State()
	if err != nil {
		return nil, err
	}
	
	// <create and return the client>
}
```

Then TaskStateLister can simply call `MakeClientFn` on-demand, whenever it needs a client. And similarly the the TaskState type could switch to using `MakeClientFn`.

Just one idea. What do you think?

config/types.go
Ganeshrockz: Yeah. That makes more sense. I just wanted to be extra cautious on making sure the controller is truly backward compatible. Looks like the node name should be enough to ensure compat.

controller/resource.go
Ganeshrockz: Yeah. This makes sense and looks a lot more cleaner in terms of responsibilities. Thanks for the feedback

cthain (commented):
Looking good! Just a couple of minor comments on the code changes. I'm reviewing the tests now.

controller/resource.go
```diff
@@ -221,20 +244,53 @@ func (s TaskStateLister) fetchACLState() (map[TaskID]*TaskState, error) {
...
 }
 
+func (s TaskStateLister) fetchServiceStateForTasks(consulClient *api.Client) (map[TaskID]*TaskState, error) {
+	opts := &api.QueryOptions{Partition: s.Partition}
+	if s.Partition != "" {
+		opts.Namespace = "*" // wildcard to fetch services from all namespaces
+	}
+	services, _, err := consulClient.Catalog().NodeServiceList(s.ClusterARN, opts)
+	if err != nil {
+		return nil, fmt.Errorf("fetching list of services for the given node %s: %w", s.ClusterARN, err)
```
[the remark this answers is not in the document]
Ganeshrockz: `ClusterARN` is the node name in the new architecture. We register it as a synthetic node to keep track of services.

controller/resource.go
```diff
@@ -221,20 +244,53 @@ func (s TaskStateLister) fetchACLState() (map[TaskID]*TaskState, error) {
...
+	}
+
+	var result error
+	serviceState := make(map[TaskID]*TaskState)
+	for _, service := range services.Services {
+		taskID, err := getTaskIDFromServiceMeta(service)
+		if err != nil {
+			s.Log.Error(err.Error())
```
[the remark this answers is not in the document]
Ganeshrockz: Makes sense

controller/resource.go
```diff
@@ -432,23 +488,51 @@ type TaskState struct {
 	ECSTaskFound bool
 	// ACLTokens are the Consul ACL tokens found for this task id.
 	ACLTokens []*api.ACLTokenListEntry
+	// Consul Service associated with this ECS task
+	Service *api.AgentService
```
pglass, line 492: We should have two service instances per task: one for the service itself, and one for the sidecar proxy.

controller/resource.go
```diff
@@ -457,15 +541,51 @@ func (t *TaskState) Delete() error {
...
 }
 
+// ID returns the taskID of the resource
+func (t *TaskState) ID() TaskID {
+	return t.TaskID
+}
+
+// IsACLNotFoundError returns true if the ACL is not found.
+func IsACLNotFoundError(err error) bool {
+	return err != nil && strings.Contains(err.Error(), "Unexpected response code: 403 (ACL not found)")
```
pglass, line 577: minor: We may want to change this check to just `strings.Contains(err.Error(), "ACL not found")`, [similar to what Consul has defined](https://github.com/hashicorp/consul/blob/efe981637b44d9e154474de2ba5ca6e6ecb4ca8d/acl/errors.go#L43-L47).

There were some ACL response code and error message changes in Consul 1.15, so I'm suspicious of comparing with a longer string here.

Non-blocking because you haven't changed this, but something to keep in mind.

pglass (commented):
Looking good! Thanks for reworking the code 🙏 

We just need the controller to handle deregistering both the service instance and the sidecar proxy instance for a task.

controller/resource.go
Ganeshrockz: My bad. Missed this one. Thanks for pointing out

PR #154, opened 2023-07-14
Refactor TLS inputs in ECS_CONFIG_JSON

Ganeshrockz:
- Refactored the existing  TLS settings present in `consulServers` to protocol specific blocks

```json
{
  "consulServers": {
    "defaults": {   ## Common/default TLS settings for both http and grpc
      "caCertFile": "",
      "tls": true, // Defaults to true
      "tlsServerName": ""
    },
    "http": {   ## HTTP-only settings (overrides `defaults` for http)
      "caCertFile":"",
      "port": 8501, // Defaults to 8501
      "https": true, // Defaults to true
      "tls": null,
      "tlsServerName": ""
    },
    "grpc": {  ## GRPC-specific settings (overrides `defaults` for gRPC)
      "caCertFile": "",
      "port":  8503, // Defaults to 8503
      "tls": null,
      "tlsServerName": ""
    }
  }
}
```

## How I've tested this PR:

Unit tests

## Checklist:
- [X] Tests added
- [ ] CHANGELOG entry added


internal/dataplane/dataplane_config.go
```diff
@@ -45,10 +45,11 @@ func (i *GetDataplaneConfigJSONInput) GetDataplaneConfigJSON() ([]byte, error) {
 		},
 	}
 
-	if i.ConsulServerConfig.EnableTLS {
+	grpcTLSSettings := i.ConsulServerConfig.GetGRPCTLSSettings()
+	if grpcTLSSettings.Enabled {
 		cfg.Consul.TLS = &TLSConfig{
-			GRPCCACertPath: i.ConsulServerConfig.CACertFile,
-			TLSServerName:  i.ConsulServerConfig.TLSServerName,
+			GRPCCACertPath: grpcTLSSettings.CaCertFile,
```
Ganeshrockz, line 51: @pglass It just occurred to me that dataplane always accepts a file path for the CA cert and does not rely directly on a `CA_CERT_PEM` env variable. On the other hand, control plane configures the TLS settings with either a file or a env variable. Couple of options to proceed here

1. Remove ability to pass the cert PEM as a environment variable and only accept a ca cert file path in ECS_CONFIG_JSON.
2. Control plane writes the PEM to a shared volume (if a file input is not provided) and passes the path it to the dataplane as part of the dataplane config json.

I prefer option (2). Let me know your thoughts.

pglass: I agree. 

I don't think we should remove support for the environment variable because that is how values are securely passed in from Secrets Manager (or Parameter Store[1]) to ECS tasks. There is not a similar native way in AWS to pass values into via _files_ from Secrets Manager, so we need to keep the environment variables for now. (Although the CA cert in question here is technically public / non-sensitive)

https://github.com/hashicorp/terraform-aws-consul-ecs/blob/7440b1a9b1e11769d44232b0475024dccb9bcbd8/modules/mesh-task/main.tf#L259-L270

https://github.com/hashicorp/terraform-aws-consul-ecs/blob/7440b1a9b1e11769d44232b0475024dccb9bcbd8/modules/acl-controller/main.tf#L49-L53

Writing the CA cert to a shared value makes sense, and we can do the same for any other files that are passed to Consul Dataplane (such as the optional client cert/key and prometheus cert files)

---

1. Consul ECS doesn't support Parameter Store right now, but it could and should support it.

Ganeshrockz: I can take this up in a separate PR

pglass (approved):
This is great. I really like this config structure!

PR #156, opened 2023-07-18
Miscellaneous changes for dataplane architecture

Ganeshrockz:
- Added ability to write the CA cert PEM to the shared volume because dataplane only accepts a file path for the same.
- Removed XDS Bind Port so that the dataplane can decided what port to pick.
- Users would pass a bootstrap token to the controller. In such cases, asking logging in to consul via the IAM auth method might not be needed. Added ability to pass the bootstrap token as a static credential to the server connection manager's discovery credentials.

## How I've tested this PR:

Unit tests

## Checklist:
- [X] Tests added
- [ ] CHANGELOG entry added

### DIFF
--- a/internal/dataplane/dataplane_config.go
+++ b/internal/dataplane/dataplane_config.go
@@ -22,6 +22,9 @@ type GetDataplaneConfigJSONInput struct {
 	// If empty, credential details are not populated in the resulting
 	// dataplane config JSON.
 	ConsulToken string
+
+	// Path of the CA cert file for Consul server's RPC interface
+	CACertFile string
 }
 
 // GetDataplaneConfigJSON returns back a configuration JSON which
@@ -41,14 +44,18 @@ func (i *GetDataplaneConfigJSONInput) GetDataplaneConfigJSON() ([]byte, error) {
 		},
 		XDSServer: XDSServerConfig{
 			Address: localhostAddr,
-			Port:    20000,
 		},
+	}
+
+	cfg.Consul.TLS = &TLSConfig{
+		Disabled: true,
 	}
 
 	grpcTLSSettings := i.ConsulServerConfig.GetGRPCTLSSettings()
 	if grpcTLSSettings.Enabled {
 		cfg.Consul.TLS = &TLSConfig{
-			GRPCCACertPath: grpcTLSSettings.CaCertFile,
+			Disabled:       false,
+			GRPCCACertPath: i.CACertFile,
 			TLSServerName:  grpcTLSSettings.TLSServerName,
 		}
 	}
--- a/subcommand/control-plane/command.go
+++ b/subcommand/control-plane/command.go
@@ -67,6 +67,7 @@ type Command struct {
 
 const (
 	dataplaneConfigFileName = "consul-dataplane.json"
+	caCertFileName          = "mesh-task-consul-ca-cert.pem"
 
 	defaultHealthCheckBindAddr = "127.0.0.1"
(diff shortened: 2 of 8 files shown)


internal/dataplane/dataplane_config.go
```diff
@@ -41,14 +44,18 @@ func (i *GetDataplaneConfigJSONInput) GetDataplaneConfigJSON() ([]byte, error) {
 		},
 		XDSServer: XDSServerConfig{
 			Address: localhostAddr,
-			Port:    20000,
```
Ganeshrockz, line 44: Envoy complains that 20000 is already in use whenever it starts in the current state. It looks like dataplane picks a free port if we don't set it from the control plane.

subcommand/controller/command.go
```diff
@@ -103,6 +99,12 @@ func (c *Command) run() error {
 	// Set up ECS client.
 	ecsClient := ecs.New(clientSession)
 
+	// Token containing `acl:write`, `operator:write` and `node:write` privileges against the server
+	token := config.GetConsulToken()
+	if token != "" {
+		return fmt.Errorf("CONSUL_HTTP_TOKEN should be non empty")
```
pglass, line 105: I don't think we should make the token mandatory (unfortunately). If Consul does not have ACLs enabled, then the controller does not need an ACL token to run. It would need to deregister agentless services, but would not need to cleanup ACL tokens.

subcommand/control-plane/command.go
```diff
@@ -67,6 +67,7 @@ type Command struct {
 
 const (
 	dataplaneConfigFileName = "consul-dataplane.json"
+	caCertFileName          = "mesh-task-consul-ca-cert.pem"
```
pglass, line 70: nit: Naming suggestion

```suggestion
	caCertFileName          = "consul-grpc-ca-cert.pem"
```

internal/dataplane/dataplane_config.go
pglass: 🤔 I'm wondering if this is going to work in that case:

* Consul Dataplane [starts an xDS server/proxy](https://github.com/hashicorp/consul-dataplane/blob/9d470c739e3eeb0a4555c452482b50acb376227c/pkg/consuldp/xds.go#L56)

Then it configures an Envoy cluster to point at that xDS server:

* https://github.com/hashicorp/consul-dataplane/blob/9d470c739e3eeb0a4555c452482b50acb376227c/pkg/consuldp/bootstrap.go#L59
* https://github.com/hashicorp/consul-dataplane/blob/9d470c739e3eeb0a4555c452482b50acb376227c/internal/bootstrap/bootstrap_tpl.go#L240

I think if the Envoy `socket_address.port_value = 0` then it wouldn't point at the xDS server started by Consul Dataplane that is running on a randomly selected port.

Do you know if it will work? I might be missing something.

Ganeshrockz: It does work because `socker_address.port_value` gets the value from the XDS listener address and not the `xdsServer.BindPort` present in dataplane config.

https://github.com/hashicorp/consul-dataplane/blob/9d470c739e3eeb0a4555c452482b50acb376227c/pkg/consuldp/bootstrap.go#L84

subcommand/controller/command.go
Ganeshrockz: Cool. In that case, we could make this token optional. If a token is present we use it to login. If not, we login via the auth method configuration provided via the `consulLogin` field.

pglass (approved):

PR #160, opened 2023-08-09
Added go-discover binary to the ECS image

Ganeshrockz:
- Adds the `go-discover` CLI binary to the `consul-ecs` image. This will be used by the new control plane to discover servers when `consulServers.hosts` is of the form

```
exec /bin/discover addrs provider="aws" region="us-east-1" ...
```

## How I've tested this PR:

Rely on CI

## Checklist:
- [ ] Tests added N/A
- [ ] CHANGELOG entry added

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,7 +8,7 @@
 # when building.
 
 # go-discover builds the discover binary
-FROM golang:1.20.4-alpine as go-discover
+FROM golang:1.20.7-alpine as go-discover
 RUN CGO_ENABLED=0 go install github.com/hashicorp/go-discover/cmd/discover@214571b6a5309addf3db7775f4ee8cf4d264fd5f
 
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -7,6 +7,10 @@
 # Every target has a BIN_NAME argument that must be provided via --build-arg=BIN_NAME=<name>
 # when building.
 
+# go-discover builds the discover binary
+FROM golang:1.20.4-alpine as go-discover
```
cthain, line 11: The `golang:1.20.7-alpine` image is available and the latest release. We should probably use that one.

```suggestion
FROM golang:1.20.7-alpine as go-discover
```

cthain (approved):
Left a minor, non-blocking suggestion to bump the golang image version.

pglass (approved):

PR #167, opened 2023-08-31
Add locality params to service registration

Ganeshrockz:
- Add locality related params to catalog service registration.
- Did a `go get github.com/hashicorp/consul/api@main` and `go mod tidy` to point to the latest API definitions in Consul's main branch.
- Removed Consul 1.15 and 1.14 binaries from the test matrix because the unit tests will fail for those versions since the parameters needed for setting service locality won't be available in those versions.

## How I've tested this PR:

Unit tests

## Checklist:
- [X] Tests added
- [X] CHANGELOG entry added


awsutil/awsutil.go
```diff
@@ -173,3 +178,7 @@ func NewSession(meta ECSTaskMeta, userAgentCaller string) (*session.Session, err
 	}
 	return clientSession, nil
 }
+
+func GetAWSRegion() string {
+	return os.Getenv(AWSRegionEnvVar)
```
absolutelightning, line 183: I think taskARN has region can we use that?

Ganeshrockz: Region is embedded in the taskARN which brings in additional challenges where we have to parse and extract it out. `AWS_REGION` is something that ECS injects automatically into the container.

absolutelightning (approved):

CHANGELOG.md
```diff
@@ -31,6 +31,9 @@ BREAKING CHANGES
    - Add the `meshGateway.healthCheckPort` field which can be hit to determine Envoy's readiness.
 * Add the [go-discover](https://github.com/hashicorp/go-discover) binary to the Consul ECS image to better support [cloud auto-join](https://developer.hashicorp.com/consul/docs/install/cloud-auto-join).[[GH-160](https://github.com/hashicorp/consul-ecs/pull/160)]
 
+FEATURES
+* Sets the `AWS_REGION` container environment variable and `AvailabilityZone` attribute of an ECS task meta JSON as locality parameters in Consul service registration's payload. These parameters will only be used to perform locality aware routing only for Enterprise Consul installations. [[GH-167](https://github.com/hashicorp/consul-ecs/pull/167)]
```
cthain, line 35: Suggest changing the wording here to make it more clear that we are using the properties from the ECS task to set the locality params:

```suggestion
* Use the `AWS_REGION` container environment variable and `AvailabilityZone` attribute of an ECS task meta JSON to set the locality parameters in Consul service and proxy registrations. These parameters are used to perform locality aware routing for Consul Enterprise installations. [[GH-167](https://github.com/hashicorp/consul-ecs/pull/167)]
```

subcommand/control-plane/command_test.go
```diff
@@ -439,6 +447,10 @@ func TestRun(t *testing.T) {
 				},
 				Partition: expectedPartition,
 				Namespace: expectedNamespace,
+				ServiceLocality: &api.Locality{
+					Region: testRegion,
+					Zone:   testZone,
+				},
```
cthain, line 453: non-blocking: Do you think it's worth testing the case where we return a `nil` locality? Given that the `AWS_REGION` will always be provided by the ECS environment this would really just be a code coverage case.

cthain (approved):
Left some non-blocking comments/suggestions for your consideration.

PR #170, opened 2023-09-12
Add support to add Consul dataplane as a nameserver

Ganeshrockz:
- Adds a package that reconfigures the `/etc/resolv.conf` file to add localhost as a nameserver. This enables us to define upstreams as Consul DNS names in the ECS task definition. With Tproxy, the request gets transparently routed to Consul dataplane's DNS interface which then proxies the request to Consul DNS server.

## How I've tested this PR:

CI

👀 

## Checklist:
- [X] Tests added
- [ ] CHANGELOG entry added (This PR is raised against a feature branch)

### DIFF
--- a/internal/dns/dns.go
+++ b/internal/dns/dns.go
@@ -1,0 +1,97 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
+
+package dns
+
+import (
+	"bytes"
+	"fmt"
+	"os"
+	"strings"
+
+	"github.com/miekg/dns"
+)
+
+const (
+	// Defaults taken from /etc/resolv.conf man page
+	defaultDNSOptionNdots    = 1
+	defaultDNSOptionTimeout  = 5
+	defaultDNSOptionAttempts = 2
+
+	defaultEtcResolvConfFile   = "/etc/resolv.conf"
+	consulDataplaneDNSBindHost = "127.0.0.1"
+)
+
+type ConfigureConsulDNSInput struct {
+	// Used only for unit tests
+	etcResolvConfFile string
+}
+
+// ConfigureConsulDNS reconstructs the /etc/resolv.conf file by setting the
+// consul-dataplane's DNS server (i.e. localhost) as the first nameserver in the list.
+func (i *ConfigureConsulDNSInput) ConfigureConsulDNS() error {
+	etcResolvConfFile := defaultEtcResolvConfFile
+	if i.etcResolvConfFile != "" {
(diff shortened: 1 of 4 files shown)


internal/dns/dns.go
```diff
@@ -0,0 +1,97 @@
...
+	"github.com/miekg/dns"
+)
+
+const (
+	// Defaults taken from /etc/resolv.conf man page
+	defaultDNSOptionNdots    = 1
+	defaultDNSOptionTimeout  = 5
+	defaultDNSOptionAttempts = 2
+
+	defaultEtcResolvConfFile   = "/etc/resolv.conf"
```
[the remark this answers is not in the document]
Ganeshrockz: Yes. This is only for Linux for now.

absolutelightning:
One last doubt that I have before approving. Basically you are reading from a file and the writing/overriding same file right?
But some contents from that file is not read like - sortlist and couple of other options - https://man7.org/linux/man-pages/man5/resolv.conf.5.html. Should we read the file and check for nameserver: "127.0.0.1" and if not present just append this to the file? @Ganeshrockz

internal/dns/dns.go
```diff
@@ -0,0 +1,97 @@
...
+	}
+
+	if cfg == nil {
+		return fmt.Errorf("failed to fetch DNS config")
+	}
+
+	options := constructDNSOpts(cfg)
+
+	nameservers := []string{consulDataplaneDNSBindHost}
+	nameservers = append(nameservers, cfg.Servers...)
```
absolutelightning, line 50: what if nameserver `consulDataplaneDNSBindHost` is already present?

Ganeshrockz: It is not expected for localhost to present as a `nameserver`. By default ECS tasks point to a global nameserver for resolving IPs. Hypothetically if we were to encounter such a case, it should just be a duplicate nameserver entry and the lookup should fallback to the subsequent entries in the list. The aim here is to just move the localhost nameserver to the top of the list to make sure we fallback from there on.

Ganeshrockz:
> But some contents from that file is not read like - sortlist and couple of other options - https://man7.org/linux/man-pages/man5/resolv.conf.5.html

I had a similar doubt while authoring the RFC and took inspiration from what Kubernetes does. We only focus on these three primary fields and options while reconfiguring the file. I have also verified that tasks in the `awsvpc` networking mode only have these three fields. We could keep track of this and reiterate if someone sees issues while dealing with other DNS options. Thoughts?

absolutelightning (approved):

PR #173, opened 2023-09-13
Added package for supporting traffic redirection

Ganeshrockz:
- Adds the `redirecttraffic` package that will support traffic redirection for transparent proxying requests in ECS.
- Makes use of the `iptables` Consul SDK to configure the rules
- Most of the exclusion scenarios are covered in detail in the RFC.
 

## How I've tested this PR:
CI

👀 

## Checklist:
- [X] Tests added
- [ ] CHANGELOG entry added n/a

### DIFF
--- a/internal/redirecttraffic/redirect_traffic.go
+++ b/internal/redirecttraffic/redirect_traffic.go
@@ -66,7 +66,7 @@ func WithIPTablesProvider(provider iptables.Provider) TrafficRedirectionOpts {
 }
 
 func New(cfg *config.Config, proxySvc *api.AgentService, consulServerAddress, clusterARN string, proxyHealthCheckPort int, opts ...TrafficRedirectionOpts) TrafficRedirectionProvider {
-	input := &TrafficRedirectionCfg{
+	trafficRedirectionCfg := &TrafficRedirectionCfg{
 		ProxySvc:             proxySvc,
 		ConsulServerAddress:  consulServerAddress,
 		ClusterARN:           clusterARN,
@@ -79,10 +79,10 @@ func New(cfg *config.Config, proxySvc *api.AgentService, consulServerAddress, cl
 	}
 
 	for _, opt := range opts {
-		opt(input)
+		opt(trafficRedirectionCfg)
 	}
 
-	return input
+	return trafficRedirectionCfg
 }
 
 // applyTrafficRedirectionRules creates and applies traffic redirection rules with
@@ -176,8 +176,7 @@ func (c *TrafficRedirectionCfg) Apply() error {
 	if err != nil {
 		return err
 	}
-	c.iptablesCfg.ExcludeOutboundCIDRs = append(c.iptablesCfg.ExcludeOutboundCIDRs, fmt.Sprintf("%s/32", parsedURL.Host))
-	c.iptablesCfg.ExcludeOutboundCIDRs = append(c.iptablesCfg.ExcludeOutboundCIDRs, fmt.Sprintf("%s/32", c.ConsulServerAddress))
+	c.iptablesCfg.ExcludeOutboundCIDRs = append(c.iptablesCfg.ExcludeOutboundCIDRs, fmt.Sprintf("%s/32", parsedURL.Host), fmt.Sprintf("%s/32", c.ConsulServerAddress))
 
 	// UIDs
 	c.iptablesCfg.ExcludeUIDs = append(c.iptablesCfg.ExcludeUIDs, c.ExcludeUIDs...)
--- a/internal/redirecttraffic/redirect_traffic.go
+++ b/internal/redirecttraffic/redirect_traffic.go
@@ -1,0 +1,205 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
+
+package redirecttraffic
+
+import (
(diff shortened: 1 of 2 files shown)


internal/redirecttraffic/redirect_traffic.go
```diff
@@ -0,0 +1,205 @@
...
+		}
+
+		// Exclude Envoy's ready bind port that indicate envoy's readiness
+		c.iptablesCfg.ExcludeInboundPorts = append(c.iptablesCfg.ExcludeInboundPorts, strconv.Itoa(c.ProxyHealthCheckPort))
+
+		// Exclude envoy_prometheus_bind_addr port from inbound redirection rules.
+		if trCfg.PrometheusBindAddr != "" {
+			_, port, err := net.SplitHostPort(trCfg.PrometheusBindAddr)
+			if err != nil {
+				return fmt.Errorf("failed parsing host and port from envoy_prometheus_bind_addr: %w", err)
```
absolutelightning, line 142: should we log `PrometheusBindAddr` or it will be logged by itself?

internal/redirecttraffic/redirect_traffic.go
```diff
@@ -0,0 +1,205 @@
...
+	// Outbound CIDRs
+	c.iptablesCfg.ExcludeOutboundCIDRs = append(c.iptablesCfg.ExcludeOutboundCIDRs, c.ExcludeOutboundCIDRs...)
+
+	// Exclude TaskMeta and Consul Server IPs from outbound traffic redirection
+	parsedURL, err := url.Parse(os.Getenv(awsutil.ECSMetadataURIEnvVar))
+	if err != nil {
+		return err
+	}
+	c.iptablesCfg.ExcludeOutboundCIDRs = append(c.iptablesCfg.ExcludeOutboundCIDRs, fmt.Sprintf("%s/32", parsedURL.Host))
+	c.iptablesCfg.ExcludeOutboundCIDRs = append(c.iptablesCfg.ExcludeOutboundCIDRs, fmt.Sprintf("%s/32", c.ConsulServerAddress))
```
absolutelightning, line 180: Should we merge the above 2 lines into one?

internal/redirecttraffic/redirect_traffic.go
```diff
@@ -0,0 +1,205 @@
...
+//	ExcludeOutboundCIDRs: TaskMeta IP, Consul Server IP and `transparentProxy.excludeOutboundCIDRs` in CONSUL_ECS_CONFIG_JSON
+//	ExcludeUIDs: `transparentProxy.excludeUIDs` in CONSUL_ECS_CONFIG_JSON
+func (c *TrafficRedirectionCfg) Apply() error {
+	if c.ProxySvc == nil {
+		return fmt.Errorf("proxy service details are required to enable traffic redirection")
+	}
+
+	// Decode proxy's opaque config
+	var trCfg trafficRedirectProxyConfig
+	if err := mapstructure.WeakDecode(c.ProxySvc.Proxy.Config, &trCfg); err != nil {
```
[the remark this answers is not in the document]
Ganeshrockz: It can't be `nil` for a proxy service. We could may be add this check when passing the proxySvc to the `Apply` function in the caller. Those changes will be added in a followup PR.

internal/redirecttraffic/redirect_traffic.go
```diff
@@ -0,0 +1,205 @@
...
+		ProxyHealthCheckPort: proxyHealthCheckPort,
+		EnableConsulDNS:      cfg.ConsulDNSEnabled(),
+		ExcludeInboundPorts:  cfg.TransparentProxy.ExcludeInboundPorts,
+		ExcludeOutboundPorts: cfg.TransparentProxy.ExcludeOutboundPorts,
+		ExcludeOutboundCIDRs: cfg.TransparentProxy.ExcludeOutboundCIDRs,
+		ExcludeUIDs:          cfg.TransparentProxy.ExcludeUIDs,
+	}
+
+	for _, opt := range opts {
+		opt(input)
```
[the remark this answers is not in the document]
Ganeshrockz: This is a way to provide optional function arguments in Go. When someone invokes this function in [this](https://github.com/hashicorp/consul-ecs/blob/7eba201cc0b4852c57324ad611a224956fd2a6e8/internal/redirecttraffic/redirect_traffic_test.go#L210) manner, the corresponding `WithIPTablesProvider` handler gets called which alters the input config to add the passed provider.

This loop is meant to add multiple such optional parameters in the future if needed. For example, let's say there is a new optional field `foo` that gets added in the struct, it can be populated with the help of a function that looks something like

```golang
func WithFoo(foo string) TrafficRedirectionOpts {
	return func(c *TrafficRedirectionCfg) {
		c.foo = foo
	}
}
```

and the caller would do something like

```golang
trafficRedirectionProvider.Apply(
.....
.....
WithIPTablesProvider(customIPTablesProvider),
WithFoo("foo"),
)
```

This loop will iterator through the `opts` input (which are nothing but functions) and start applying them with `opt(input)`

internal/redirecttraffic/redirect_traffic.go
Ganeshrockz: Address gets logged automatically in the `err` variable.

absolutelightning (approved):

PR #176, opened 2023-09-22
Add proxy.localServiceAddress to enable configuration of Envoy local_app

oulman:
- Adds a `proxy.localServiceAddress` field to configure Envoy's local_app endpoint address

## How I've tested this PR:

1. Deployed the [cluster-peering example](https://github.com/hashicorp/terraform-aws-consul-ecs/tree/main/examples/cluster-peering) example with a image built from this PR.
2. Modified CONSUL_ECS_CONFIG_JSON to include a `proxy.localServiceAddress` field set to the IP of the Fargate task.
3. Verified that the consul-dataplane containers Envoy config specified the task ip in the local_app cluster rather than localhost.

```
      "load_assignment": {
       "cluster_name": "local_app",
       "endpoints": [
        {
         "lb_endpoints": [
          {
           "endpoint": {
            "address": {
             "socket_address": {
              "address": "10.0.1.140",
              "port_value": 7933
             }
            }
           }
          }
         ]
        }
```

## Checklist:
- [x] Tests added
- [x] CHANGELOG entry added

### DIFF
--- a/config/schema.json
+++ b/config/schema.json
@@ -238,6 +238,10 @@
           "description": "Object value that specifies an opaque JSON configuration. The JSON is stored and returned along with the service instance when called from the API.",
           "type": ["object", "null"]
         },
+        "localServiceAddress": {
+          "description": "The IP address or hostname that Envoy should use to connect to the local service. Defaults to localhost.",
+          "type": ["string", "null"]
+        },
         "publicListenerPort": {
           "description": "The public listener port for Envoy used for service-to-service communication. Defaults to 20000.",
           "type": ["integer", "null"]
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@ BREAKING CHANGES
    - Add the `proxy.healthCheckPort` field which can be hit to determine Envoy's readiness.
    - Add the `proxy.upstreams.destinationPeer` field to enable the proxy to hit upstreams present in peer Consul clusters.
    - Add the `meshGateway.healthCheckPort` field which can be hit to determine Envoy's readiness.
+   - Add the `proxy.localServiceAddress` field to configure Envoy to use a different address for the local service.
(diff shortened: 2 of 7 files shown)


oulman:
@Ganeshrockz I wasn't sure if you wanted to switch the proxy tests to be table driven or not to make them easier to manage.  I went ahead and just [added a second set](https://github.com/hashicorp/consul-ecs/pull/176/files#diff-7ee5f56a3c711afa50ed603205450b67cd39fdafcace9462830645cba3580514R25-R28) to make sure the tests worked.

config/schema.json
```diff
@@ -238,6 +238,10 @@
           "description": "Object value that specifies an opaque JSON configuration. The JSON is stored and returned along with the service instance when called from the API.",
           "type": ["object", "null"]
         },
+        "localServiceAddress": {
```
Ganeshrockz, line 241: Can you add some tests for this new field? Existing tests for parsing the schema can be found here https://github.com/hashicorp/consul-ecs/blob/main/config/validate_test.go

oulman: @Ganeshrockz Thanks and done! I added them to the existing `expectedExtensiveConfig` based tests since the omitted value should be covered by other tests. Please let me know if you would like me to rework that into its own test. Thanks!

Ganeshrockz (approved):

config/schema.json
Ganeshrockz: This looks good. Thanks for adding those tests.

Ganeshrockz:
@oulman I am planning to cut off the release candidate branch for releasing the new version of Consul ECS. Is this change needed there as well?

oulman:
@Ganeshrockz Yeah, if we could that would be great! I assume the -rc version would make it into a public registry and we could stop using our custom build.

PR #207, opened 2023-12-20
Offload tons of responsibilities from control-plane

Ganeshrockz:

One of many PRs that involved remodelling Consul on ECS to support TProxy. This PR focusses on removing the following functionalities from the control-plane container.

- Made it a short lived container
- Removed custom health check server
- Removed dataplane monitor to facilitate graceful shutdown
- Removed reconciliation loop that sync back ECS health checks into Consul.
- Removed logic to deregisterServiceAndProxy and Consul logout from control plane.
- Lots of code related to syncing ECS health checks into Consul
- Bunch of test code that is related to ☝️ 

In the new architecture, control plane will only do the following

- Perform a consul login and write the token to a shared volume
- Register server and proxy
- Write the ECS binary to a shared volume
- Write dataplane config to a shared volume

All this logic that got removed will be added back as part of a new `health-sync` command in the same binary. `Control-plane` will be renamed to mesh-init in the next PR

## How I've tested this PR:

CI

## Checklist:
- [X] Tests added (Rather it should be tests removed :))
- [ ] CHANGELOG entry added N/A. This is only against a feature branch for now.

### DIFF
--- a/subcommand/control-plane/checks.go
+++ b/subcommand/control-plane/checks.go
@@ -5,13 +5,9 @@ package controlplane
 
 import (
 	"fmt"
-	"time"
 
-	"github.com/aws/aws-sdk-go/service/ecs"
-	"github.com/hashicorp/consul-ecs/awsutil"
 	"github.com/hashicorp/consul-ecs/config"
 	"github.com/hashicorp/consul/api"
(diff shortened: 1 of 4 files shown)


absolutelightning (approved):
LGTM 🎉

PR #210, opened 2023-12-22
Add health-sync command to ECS

Ganeshrockz:
- Adds a new command `health-sync` to the command factory.
- Adds back most of the code removed in https://github.com/hashicorp/consul-ecs/pull/207 but in the `healthsync` package.
- Health sync container now does the following
   - Performs Consul login with the connection manager library
   - Sets up the Consul client
   - Accumulates all the health checks from the catalog for the service/proxy registered by `mesh-init`
   - Enters into a long running reconciliation loop where it (Most of the code is a copy paste of the reconciliation logic that was previously present in Control plane removed in https://github.com/hashicorp/consul-ecs/pull/207)
      - Periodically syncs back ECS container health checks into Consul. 
      - Marks all the checks as critical on receiving SIGTERM
      - Implements a server watch where it listens for changes to the Consul servers and reconfigures the Consul client when such a change happens
      - Implements graceful shutdown where it waits for Consul dataplane to terminate. Once dataplane terminates, it deregisters the service and proxy and finally performs a Consul logout.

## How I've tested this PR:

- Tests were inspired from the code removed in https://github.com/hashicorp/consul-ecs/pull/207 and added here according to the changes.

## Checklist:
- [X] Tests added
- [ ] CHANGELOG entry added

### DIFF
--- a/subcommand/health-sync/checks.go
+++ b/subcommand/health-sync/checks.go
@@ -1,0 +1,255 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
+
+package healthsync
+
+import (
+	"fmt"
+
+	"github.com/aws/aws-sdk-go/service/ecs"
+	"github.com/hashicorp/consul-ecs/awsutil"
+	"github.com/hashicorp/consul-ecs/config"
+	"github.com/hashicorp/consul/api"
+	"github.com/hashicorp/go-multierror"
+)
+
+// fetchHealthChecks fetches the Consul health checks for both the service
+// and proxy registrations
+func (c *Command) fetchHealthChecks(consulClient *api.Client, taskMeta awsutil.ECSTaskMeta) (map[string]*api.HealthCheck, error) {
+	serviceName := c.constructServiceName(taskMeta.Family)
+	serviceID := makeServiceID(serviceName, taskMeta.TaskID())
+	proxySvcID, proxySvcName := makeProxySvcIDAndName(serviceID, serviceName)
+
+	healthCheckMap := make(map[string]*api.HealthCheck)
+	var queryOpts *api.QueryOptions
+	if c.config.IsGateway() {
+		queryOpts = &api.QueryOptions{
+			Namespace: c.config.Gateway.Namespace,
+			Partition: c.config.Gateway.Partition,
+		}
+	} else {
+		queryOpts = &api.QueryOptions{
+			Namespace: c.config.Service.Namespace,
+			Partition: c.config.Service.Partition,
+		}
+	}
+
+	checks, err := getServiceHealthChecks(consulClient, serviceName, serviceID, queryOpts)
+	if err != nil {
+		return nil, err
+	}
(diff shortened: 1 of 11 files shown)


subcommand/health-sync/checks.go
```diff
@@ -0,0 +1,255 @@
...
+	"github.com/aws/aws-sdk-go/service/ecs"
+	"github.com/hashicorp/consul-ecs/awsutil"
+	"github.com/hashicorp/consul-ecs/config"
+	"github.com/hashicorp/consul/api"
+	"github.com/hashicorp/go-multierror"
+)
+
+// fetchHealthChecks fetches the Consul health checks for both the service
+// and proxy registrations
+func (c *Command) fetchHealthChecks(consulClient *api.Client, taskMeta awsutil.ECSTaskMeta) (map[string]*api.HealthCheck, error) {
```
Ganeshrockz, line 18: This is a newly added function. Others in this PR are a copy paste of code removed in https://github.com/hashicorp/consul-ecs/pull/207

subcommand/health-sync/command.go
```diff
@@ -0,0 +1,297 @@
...
+	if err != nil {
+		return fmt.Errorf("unable to fetch checks before the reconciliation loop %w", err)
+	}
+
+	if c.isTestEnv {
+		close(c.doneChan)
+		<-c.proceedChan
+	}
+
+	for {
```
Ganeshrockz, line 151: The entire reconciliation loop was just a copy paste from https://github.com/hashicorp/consul-ecs/pull/207

subcommand/health-sync/dataplane_monitor.go
```diff
@@ -0,0 +1,81 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
```
Ganeshrockz, line 2: This file is also a copy paste of the existing implementation removed in https://github.com/hashicorp/consul-ecs/pull/209

subcommand/health-sync/checks.go
```diff
@@ -0,0 +1,255 @@
...
+	serviceName := c.constructServiceName(taskMeta.Family)
+	serviceID := makeServiceID(serviceName, taskMeta.TaskID())
+	proxySvcID, proxySvcName := makeProxySvcIDAndName(serviceID, serviceName)
+
+	healthCheckMap := make(map[string]*api.HealthCheck)
```
absolutelightning, line 23: generally we are advised to use `var healthCheckMap map[string]*api.HealthCheck`

Ganeshrockz: Why is that? The zero valued initialization for a map type is not an empty map but `nil`. The behaviour is in contrast to string and bool types. We will end up with `nil` panics if we follow zero valued initialization here. Also it is safe to assume that the `healthCheckMap` will contain atleast one entry by the end of this function and so initialization doesn't won't be wasted.

absolutelightning (approved):

PR #242, opened 2024-05-01
[NET-9065] mesh-init: fix executable copy when dynamic linker is used

zalimeni:
When using dynamic linking (e.g. Cgo) during compilation, the `consul-ecs` process will be started by calling the linker to launch the actual Go executable. This causes `os.Executable()` to return the linker rather than the program itself.

When self-copying in the `mesh-init` command, check for dynamic linker use. If detected, fall back to a path-based executable resolution derived from `os.Argv[0]`, which should provide the name of the real executable file +/- its path.

- Add fallback for resolving the local executable path in dynamically-linked execution contexts such as Cgo + `linux/arm64`
- Several related doc comments and fixups for `Makefile`/CI

## How I've tested this PR:

Since this change is inherently difficult to test reliably due to platform dependence, I opted for manual testing.

Created a derivative Docker container from `consul-ecs-fips:0.8` w/ a patched binary, deployed to ECS and confirmed bug is fixed.

Before (logging changes from PR included w/o fix to demonstrate issue):
image
After:
image

👀 + 🤔 

## Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added


subcommand/mesh-init/command.go
```diff
@@ -375,6 +378,28 @@ func (c *Command) copyECSBinaryToSharedVolume() error {
...
+		}
+		pathEx, err = filepath.EvalSymlinks(pathEx)
+		if err != nil {
+			return err
+		}
+		pathEx, err = filepath.Abs(pathEx)
+		if err != nil {
+			return err
+		}
+		ex = pathEx
```
DanStough, line 400: Maybe a dumb question, but is there any risk to just doing this in all cases? That way we only have one code path.

DanStough (approved):
Non-blocking question. Great catch!

subcommand/mesh-init/command.go
zalimeni: Not dumb at all! I had the same thought. My main reasons for sticking w/ the fallback were:

- Isolate the fix to what’s broken (FIPS) to avoid regression, esp. since this bug already came from unexpected edge case behavior
- Try to lean on stdlib where we can in hopes of this eventually dying (I’m not certain but I think this could qualify as a de-facto bug in the Linux/procfs implementation of `os.Executable()` that we could report)

PR #256, opened 2024-05-15
build: update gha and fix cache error

DanStough:
- Another TSCCR GHA update
- Remove the caching step from the `test` workflow because this is now handled automatically by the `setup-go` action in a prior step.

I will add this to the other backports.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -23,18 +23,17 @@ jobs:
     steps:
     - name: Checkout
       uses: actions/checkout@0ad4b8fadaa221de15dcec353f45205ec38ea70b # v4.1.4
-    - uses: actions/setup-go@0c52d547c9bc32b1aa3301fd7a9cb496313a4491 # v5.0.0
+    - uses: actions/setup-go@cdcb36043654635271a94b9a6d1392de5bb323a7 # v5.0.1
       with:
         go-version: ${{ needs.get-go-version.outputs.go-version }}
     - name: golangci-lint
-      uses: golangci/golangci-lint-action@82d40c283aeb1f2b6595839195e95c2d6a49081b # v5.0.0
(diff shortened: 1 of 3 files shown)


.github/workflows/test.yml
```diff
@@ -56,19 +55,9 @@ jobs:
...
         go-version: ${{ needs.get-go-version.outputs.go-version }}
-
-    - uses: actions/cache@0c45773b623bea8c8e75f6c82b208c3cf94ea4f9 # v4.0.2
-      with:
-        path: |
-          ~/.cache/go-build
-          ~/go/pkg/mod
-        key: ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}
-        restore-keys: |
-          ${{ runner.os }}-go-
```
DanStough, line 70: No longer needed because `setup-go` handles this. See https://github.com/actions/cache/issues/1104#issuecomment-1598429474

curtbushko (approved):

.github/workflows/test.yml
DanStough: The CRT actions also disable caching in setup-go. Something to be said about starting with a clean machine for each build.

.github/workflows/test.yml
```diff
@@ -23,18 +23,17 @@ jobs:
...
-      uses: golangci/golangci-lint-action@82d40c283aeb1f2b6595839195e95c2d6a49081b # v5.0.0
+      uses: golangci/golangci-lint-action@a4f60bb28d35aeee14e6880718e0c85ff1882e64 # v6.0.1
       with:
         version: v1.54.2
         args: |
           --verbose
         only-new-issues: false
-        skip-pkg-cache: true
-        skip-build-cache: true
+        skip-cache: true
```
DanStough, line 36: This option is unrelated to the deprecated ones, but it is also set in `consul` because we've had problems with the lint cache previously.

PR #267, opened 2024-12-10
[NET11250]Added fix for health check update for consul-dataplane

abhishek-hashicorp:
When we update the consul-dataplane health status, we should take into account the service health status as well

- Modified the health sync for `consul-dataplane` to take into account the service health as well during health-sync checks

## How I've tested this PR:
Created an AWS ECS cluster with 2 services `frontend` and `backend`. The port for the backend health was misconfigured so that the service fails. 
**Before the fix:** 
On accessing `curl 0:19000/clusters` from the `frontend` service, we found that the health for `backend` service was showing as healthy. 
image

**After the fix:**
The `curl 0:19000/clusters` was showing 
image

Once the port was fixed, the `backend` service was again showing as healthy using the `curl 0:19000/clusters` endpoint from `frontend`
👀 
## Checklist:
- [x] Tests added
- [x] CHANGELOG entry added

### DIFF
--- a/subcommand/health-sync/checks.go
+++ b/subcommand/health-sync/checks.go
@@ -130,6 +130,7 @@ func (c *Command) syncChecks(consulClient *api.Client,
 		}
 	}
 
+	overallDataplaneHealth := 0
 	parsedContainers := make(map[string]string)
 	// iterate over parse
 	for _, container := range containersToSync {
@@ -160,15 +161,26 @@ func (c *Command) syncChecks(consulClient *api.Client,
 				currentStatuses[container.Name] = container.Health.Status
 			}
 		}
-
+		if container.Name == config.ConsulDataplaneContainerName {
+			if container.Health.Status == ecs.HealthStatusHealthy {
+				overallDataplaneHealth = 1
+			}
+		}
 	}
 
-	overallDataplaneHealthStatus := ecs.HealthStatusHealthy
-	for _, healthStatus := range parsedContainers {
-		if healthStatus != ecs.HealthStatusHealthy {
-			overallDataplaneHealthStatus = ecs.HealthStatusUnhealthy
-			break
+	for containerName, healthStatus := range parsedContainers {
+
+		if containerName != config.ConsulDataplaneContainerName {
+			currentContainerHealth := 0
+			if healthStatus == ecs.HealthStatusHealthy {
+				currentContainerHealth = 1
+			}
+			overallDataplaneHealth = overallDataplaneHealth & currentContainerHealth
 		}
+	}
+	overallDataplaneHealthStatus := ecs.HealthStatusUnhealthy
+	if overallDataplaneHealth == 1 {
+		overallDataplaneHealthStatus = ecs.HealthStatusHealthy
 	}
 
 	err = c.handleHealthForDataplaneContainer(consulClient, taskMeta.TaskID(), serviceName, clusterARN, config.ConsulDataplaneContainerName, overallDataplaneHealthStatus)
--- a/subcommand/health-sync/checks.go
+++ b/subcommand/health-sync/checks.go
@@ -166,15 +166,15 @@ func (c *Command) syncChecks(consulClient *api.Client,
 	// if dataplane container exist and healthy then proceed to checking the other containers health
 	if ok && overallDataplaneHealthStatus == ecs.HealthStatusHealthy {
 		//
-		for name, healthStatus := range parsedContainers {
+		for _, healthStatus := range parsedContainers {
 			// as soon as we find any unhealthy container, we can set the dataplane health to unhealthy
-			if healthStatus != ecs.HealthStatusHealthy && name != config.ConsulDataplaneContainerName {
+			if healthStatus != ecs.HealthStatusHealthy {
 				overallDataplaneHealthStatus = ecs.HealthStatusUnhealthy
 				break
 			}
 		}
 	} else {
-		// If no dataplane container or dataplane container is not healthy set overall health to healthy
+		// If no dataplane container or dataplane container is not healthy set overall health to unhealthy
 		overallDataplaneHealthStatus = ecs.HealthStatusUnhealthy
 	}
 
--- a/subcommand/health-sync/checks.go
+++ b/subcommand/health-sync/checks.go
@@ -184,7 +184,14 @@ func (c *Command) syncChecks(consulClient *api.Client,
 	}
 
 	err = c.handleHealthForDataplaneContainer(consulClient, taskMeta.TaskID(), serviceName, clusterARN, config.ConsulDataplaneContainerName, overallDataplaneHealthStatus)
-
+	if err != nil {
+		c.log.Warn("failed to update Consul health status", "err", err)
+	} else {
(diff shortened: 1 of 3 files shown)


subcommand/health-sync/checks.go
```diff
@@ -161,8 +161,37 @@ func (c *Command) syncChecks(consulClient *api.Client,
 				currentStatuses[container.Name] = container.Health.Status
 			}
 		}
+		if container.Name == config.ConsulDataplaneContainerName {
+			if container.Health.Status == ecs.HealthStatusHealthy {
+				overallDataplaneHealth = 1
+			}
+		}
```
dhiaayachi, line 168: If I understand right, what we are trying to achieve here is:
- if any container health is unhealthy (including dataplane) we would like to set the dataplane health to unhealthy
- otherwise dataplane health is healthy
is that right?

If so the special case in here for the specific dataplane container is not needed. Also see my suggestion below

subcommand/health-sync/checks.go
```diff
@@ -161,8 +161,37 @@ func (c *Command) syncChecks(consulClient *api.Client,
...
+			if healthStatus == ecs.HealthStatusHealthy {
+				currentContainerHealth = 1
+			}
+			overallDataplaneHealth = overallDataplaneHealth & currentContainerHealth
+		}
+	}
+	overallDataplaneHealthStatus := ecs.HealthStatusUnhealthy
+	if overallDataplaneHealth == 1 {
+		overallDataplaneHealthStatus = ecs.HealthStatusHealthy
+	}
```
dhiaayachi, line 184: I think the logic here can be simplified to the following:
```suggestion
        overallDataplaneHealthStatus := ecs.HealthStatusHealthy
	for containerName, healthStatus := range parsedContainers {
			if healthStatus != ecs.HealthStatusHealthy {
				overallDataplaneHealthStatus = ecs.HealthStatusUnhealthy
				break
			}
		}
	}
```

subcommand/health-sync/checks.go
```diff
@@ -161,8 +161,37 @@ func (c *Command) syncChecks(consulClient *api.Client,
...
 			}
 		}
+		if container.Name == config.ConsulDataplaneContainerName {
+			if container.Health.Status == ecs.HealthStatusHealthy {
+				overallDataplaneHealth = 1
+			}
+		}
 	}
 
+	for containerName, healthStatus := range parsedContainers {
```
dhiaayachi, line 171: Maybe this loop could be merged with the one above? 🤔

dhiaayachi (changes requested):
Hey @abhishek-hashicorp 
Thank you for working on this! I left a few comments for your consideration.

subcommand/health-sync/checks.go
abhishek-hashicorp: The dataplane container health isn't totally dependent on service health. There can be a case when the dataplane container itself is missing and the service containers are healthy. In that case, we cannot mark the dataplane container healthy. For the same reason we first need to iterate through all the containers and after that only, use the status of all the other container including the dataplane container's current status to decide the overall health.

subcommand/health-sync/checks.go
abhishek-hashicorp: I tried this but we need to use dataplane container health as well to decide the overall status. This results in failure of the case when daplane container itself is missing.

subcommand/health-sync/checks.go
abhishek-hashicorp: We cannot merge it since we need to know the status of all the containers first and then use that data to decide the dataplane container's overall status.

subcommand/health-sync/checks.go
dhiaayachi: That mean that `overallDataplaneHealth` is already represented by the map entry parsedContainers[config.ConsulDataplaneContainerName] so no need to add it in here.

subcommand/health-sync/checks.go
dhiaayachi: If my assumptions are right the code would look something along those lines:
```suggestion
	overallDataplaneHealthStatus, ok := parsedContainers[config.ConsulDataplaneContainerName]
	// if dataplane container exist and healthy then proceed to checking the other containers health
	if ok && overallDataplaneHealthStatus == ecs.HealthStatusHealthy {
		//
		for name, healthStatus := range parsedContainers {
			// as soon as we find a healthy container, we can set the health the healthy
			if healthStatus == ecs.HealthStatusHealthy && name != config.ConsulDataplaneContainerName {
				overallDataplaneHealthStatus = ecs.HealthStatusUnhealthy
				break
			}
		}
	} else {
		// If no dataplane container or dataplane container is not healthy set overall health to healthy
		overallDataplaneHealthStatus = ecs.HealthStatusUnhealthy
	}
```

abhishek-hashicorp: Actually, we need to check all the other containers.
It's not atleast one which needs to be healthy but if any one of the containers is unhealthy, we need to mark the dataplane as unhealthy. 
 
 Effectively what we are saying is:
 1. If dataplane is unhealthy - UNHEALTHY
 2. if dataplane is healthy - 
 - if all service containers are healthy - HEALTHY
 - if any one of the service containers are unhealthy - UNHEALTHY 
 
 Let me know if this makes sense. I had referred the tests to understand these scenarios

subcommand/health-sync/checks.go
abhishek-hashicorp: There is only 1 dataplane container. Replied on the other comment to explain this further.

subcommand/health-sync/checks.go
```diff
@@ -161,8 +160,33 @@ func (c *Command) syncChecks(consulClient *api.Client,
...
 		}
+
+	}
+	overallDataplaneHealthStatus, ok := parsedContainers[config.ConsulDataplaneContainerName]
+	// if dataplane container exist and healthy then proceed to checking the other containers health
+	if ok && overallDataplaneHealthStatus == ecs.HealthStatusHealthy {
+		//
+		for name, healthStatus := range parsedContainers {
+			// as soon as we find any unhealthy container, we can set the dataplane health to unhealthy
+			if healthStatus != ecs.HealthStatusHealthy && name != config.ConsulDataplaneContainerName {
```
dhiaayachi, line 171: I think we can simplify this:
```suggestion
			if healthStatus != ecs.HealthStatusHealthy {
```

subcommand/health-sync/checks.go
```diff
@@ -161,8 +160,33 @@ func (c *Command) syncChecks(consulClient *api.Client,
...
+				overallDataplaneHealthStatus = ecs.HealthStatusUnhealthy
+				break
+			}
+		}
+	} else {
+		// If no dataplane container or dataplane container is not healthy set overall health to healthy
```
dhiaayachi, line 177:
```suggestion
		// If no dataplane container or dataplane container is not healthy set overall health to unhealthy
```

dhiaayachi (dismissed):
Thank you for going through my comments @abhishek-hashicorp, I left 2 minor comments, but nothing blocking 🚀

srahul3 (approved):
✅ LGTM

PR #271, opened 2025-01-15
[Backport 0.9.x] Merge CRT, aws-sdk-go changes from main

abhishek-hashicorp:
Merge changes from main

## Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added

### DIFF
--- a/.release/security-scan.hcl
+++ b/.release/security-scan.hcl
@@ -13,4 +13,11 @@ binary {
 	osv          = true
 	oss_index    = false
 	nvd          = false
+	triage {
+		suppress {
+			vulnerabilites = [
+				"GO-2022-0635"
+			]
+		}
+	}
 }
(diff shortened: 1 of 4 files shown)


.release/security-scan.hcl
```diff
@@ -13,4 +13,11 @@ binary {
 	osv          = true
 	oss_index    = false
 	nvd          = false
+	triage {
+		suppress {
+			vulnerabilites = [
+				"GO-2022-0635"
+			]
+		}
+	}
```
zalimeni, line 22: ~ Was including this in addition to the dependency bump from `main` intentional/necessary? It seemed that bump was meant to appease the security scanner as wel

zalimeni (approved):
Non-blocking question, otherwise LGTM as all these changes already occurred on `main`.

.release/security-scan.hcl
abhishek-hashicorp: Upgrading the dependency didn't seem to solve the problem so seems like we need to suppress the warning. Also, since upgrading seemed harmless so I didn't revert those changes.

PR #291, opened 2025-06-12
fix: updating data-plane health status only if it is changed

Vikramarjuna:

The dataplane-container, `consul-dataplane` health is updated based on its health and all other containers health. While other containers health is updated only when there is a change, data-plane container health is updated every second.

Updated the code to update the dataplane-container status also only if it is modified from the last update.

Refactored the code such that the debug log is addd only when the update is done.

This bug was introduced when we fixed the issue to report consul-data plane health based on other consenters as well as part of the [PR](https://github.com/hashicorp/consul-ecs/pull/267)


subcommand/health-sync/checks.go
```diff
@@ -133,32 +133,30 @@ func (c *Command) syncChecks(consulClient *api.Client,
...
-			} else {
-				c.log.Info("container health check updated in Consul",
+				c.log.Debug("updating Consul check from ECS container health",
 					"name", container.Name,
 					"status", container.Health.Status,
 					"statusSince", container.Health.StatusSince,
 					"exitCode", container.Health.ExitCode,
 				)
-				currentStatuses[container.Name] = container.Health.Status
+				checkID := constructCheckID(makeServiceID(serviceName, taskMeta.TaskID()), container.Name)
```
v-rosa, line 146: @Vikramarjuna Don't we need to keep pushing the health checks to ensure the service is alive and ready to process requests? 

Consider the case where the task lost network connectivity, it can't report the status to the control plane meaning it's not healthy/capable of handling requests. How does the control plane handle such situations?

NB: Something like the above just happened our production environment, an ECS task hanged, we lost all logs after a given timestamp but the Consul control plane still considered the task as healthy and client got a bunch of 503 errors.

Vikramarjuna: Sure. I saw that this frequent update is a side effect of the mentioned PR. Based on you just described too, this updated doesn't seem to act as a heart beat. I will hold this change. Try to reproduce what you described and work on it.

PR #320, opened 2026-03-12
Fix  health-sync logic

kswap:
Fixed open github issues:
  - **Fixes #300** — *ECS Health Sync Sending Traffic to Missing Unhealthy Containers.* A health-sync container absent from ECS task metadata (e.g. application still
  starting after a deploy) was treated as if it didn't exist, so the dataplane check stayed passing and traffic reached a service whose app container wasn't ready.
  - **Fixes #309** — *health-sync sends excessive catalog registrations for dataplane container health.* The aggregated consul-dataplane status was written on every
  poll tick regardless of whether it had changed, generating ~N writes/second per task against the Consul leader.

  Additional bugs surfaced and fixed while working on the above:
  - **Double-write of dataplane status on the same tick** — when consul-dataplane was missing from task metadata, both the missing-containers loop and the aggregation
   block called `handleHealthForDataplaneContainer` with UNHEALTHY in a single iteration.
  - **Inconsistent status format** — code was mixing ECS-format (`healthy`/`unhealthy`) with Consul-format (`passing`/`critical`) strings when tracking previous
  status, which broke the delta-guard comparison in some paths.

 Refactoring (no behavior change)

  - Restructured `syncChecks` into three explicit steps: **Gather state → Compute checks → Update Consul**. Missing and present containers now feed a single
  `resolvedStatuses` map walked by one delta-guarded loop, eliminating the previous split-loop asymmetry.
  - Added `writeContainerHealth` helper that owns the dataplane-vs-other-container dispatch. Both `syncChecks` and `setChecksCritical` now route through it, removing
  duplicated logic.
  - Simplified `computeOverallDataplaneHealth` — dropped the unreachable second loop over `missingContainers` (those entries are already in `resolvedStatuses` as
  UNHEALTHY, so the first loop handles them).
  - Restored `statusSince` and `exitCode` to the per-container update log line for containers that are present in task metadata.
  
## How I've tested this PR:
- Added unit tests for all the changes done.
- Tests on local setup with logs:

 #: 1
  Test: test_service_and_proxy_registered
  Result: PASS
  What it checks: mesh-init registers both test-service and test-service-sidecar-proxy with the expected health checks
  ────────────────────────────────────────
  #: 2
  Test: test_bug1_no_redundant_writes
  Result: PASS
  What it checks: proxy check ModifyIndex stable over 10s window — confirms delta guard suppresses redundant Catalog.Register writes (#309)
  ────────────────────────────────────────
  #: 3
  Test: test_bug2_missing_container_health
  Result: PASS
  What it checks: proxy check is critical while app container is still starting; flips to passing once app is up (#300)
  ────────────────────────────────────────
  #: 4
  Test: test_status_change_triggers_write
  Result: PASS
  What it checks: real UNHEALTHY → HEALTHY transition advances ModifyIndex exactly once, then stabilises
  ────────────────────────────────────────
  #: 5
  Test: test_app_unhealthy_marks_proxy_critical
  Result: PASS
  What it checks: SIGSTOP on app → proxy critical; SIGCONT → proxy passing
  ────────────────────────────────────────
  #: 6
  Test: test_graceful_shutdown_deregisters
  Result: PASS
  What it checks: on stop-task, checks go critical within 2s and the stopped task's catalog entries deregister within 5s

  Setup: single ECS task with consul-dataplane, health-sync, mesh-init, and an nginx app container on Consul 1.22.0; consul-ecs image built from this branch's HEAD
  and pushed to ECR.

## Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added

## PCI review checklist

- [ ] I have documented a clear reason for, and description of, the change I am making.

- [ ] If applicable, I've documented a plan to revert these changes if they require more than reverting the pull request.

- [ ] If applicable, I've documented the impact of any changes to security controls.

  Examples of changes to security controls include using new access control methods, adding or removing logging pipelines, etc.


anandmukul93:
I didnt understand this testing use case 
Important : 
Container listed in healthSyncContainers - if missing from ECS metadata entirely, treating it as unhealthy(critical), not as absent/ignored.

Can you tell me why it is passing while initializing? 



if you see this PR. It has 3 distinct steps : 
https://github.com/hashicorp/consul-ecs/pull/310/changes
Gather state - fetch container health from ECS task metadata and include dataplane and missing logic to unhealthy. 
Compute checks - transform container health into Consul check statuses
Update Consul - send updates only for checks whose status has changed

Overall major changes :  
We have split paths for missing and toSync containers . it could be more organized as in the above PR. 
Why are we assuming dataplane sync to not be done or NOT . the dataplane container starts the proxy and runs a proxy XDS server for configuring envoy(gateway or sidecar) so we register the dataplane and it registers 2 checks proxy and dataplane both. If we dont maintain that it would mean that envoy is wrongly configured for a long time and can result in outdated behavior while still being healthy. so overall health should include dataplane + proxy + app container  is what i think. 

also 
subcommand/mesh-init/checks.go, constructChecks() always appends a dataplane readiness check for the given service ID (CheckID = <serviceID>-consul-dataplane)

a typical service hs 3 synced checks when you have one app container in healthSyncContainers:
service check for app container (<serviceID>-<appContainer>)
service check for dataplane (<serviceID>-consul-dataplane)
proxy check for dataplane (<proxyServiceID>-consul-dataplane)


It all starts from healthSyncContainers which container config.HealthSynccontainers which is basicallY 

```container_defs_with_depends_on = [for def in var.container_definitions :
    merge(
      def,
      {
        dependsOn = flatten(
          concat(
            lookup(def, "dependsOn", []),
            [
              {
                containerName = "consul-dataplane"
                condition     = "HEALTHY"
              }
            ]
        ))
      },
      {
        // Use the def.entryPoint, if defined. Else, use the app_entrypoint, which is null by default.
        entryPoint = lookup(def, "entryPoint", local.app_entrypoint)
        mountPoints = flatten(
          concat(
            lookup(def, "mountPoints", []),
            local.app_mountpoints,
          )
        )
      }
    )
  ]


  defaulted_check_containers or health-sync-containers = [for def in local.container_defs_with_depends_on : def.name
  
  
  # # we add dataplane over the app container definition in health-sync command. 
  
  
  1) Missing vs present still behave differently (and why that matters)
In current subcommand/health-sync/checks.go:
Missing loop writes immediately for each missing non-dataplane container (updateConsulHealthStatus call around line ~121), without checking currentStatuses[name].
Present loop first checks if container.Health.Status == currentStatuses[container.Name] { continue } (around line ~147), so unchanged present containers skip writes.
Impact:
A container that stays missing for many ticks can cause repeated Catalog.Register writes each cycle.
Present containers do not have that churn.
This asymmetry can create noisy logs and unnecessary Consul writes under startup lag / partial task metadata conditions.
It is behaviorally inconsistent with the “checks only updated when changed” comment promise in same file.


2) Dataplane special-case logic is still scattered (and why that matters)
Current dataplane handling is split across several locations in subcommand/health-sync/checks.go:
setChecksCritical dataplane branch (~74)
Missing loop skip for dataplane (~115)
Found loop skip for dataplane (~136)
Final aggregate dataplane write (~167)
Dedicated helper for dual service/proxy check updates (~208)
Impact:
Future behavior changes must touch multiple places; easy to miss one.
Harder to reason about invariants (“who owns dataplane update in this cycle?”).
 
  ```
  
    We may also look at this function that does multiple calls for different health checks. if we can club health checks in one query that would be great. 
 ```
  func (c *Command) fetchHealthChecks(consulClient *api.Client, taskMeta awsutil.ECSTaskMeta) (map[string]*api.HealthCheck, error) {
  
  ```

anandmukul93 (approved):
LGTM

anandmukul93:
[IMPORTANT]
 if the service-side write succeeds and the proxy-side write fails, handleHealthForDataplaneContainer returns the proxy error and leaves the service-side check updated in Consul and in c.checks. The next tick will diff against the now-updated currentStatuses and not retry the proxy-side write
 @kswap

kswap:
> [IMPORTANT] if the service-side write succeeds and the proxy-side write fails, handleHealthForDataplaneContainer returns the proxy error and leaves the service-side check updated in Consul and in c.checks. The next tick will diff against the now-updated currentStatuses and not retry the proxy-side write @kswap

@anandmukul93 No, the retry would happen in this case.

anandmukul93:
i think the checks already expects a slice then we can accumulate everything into one call and do that ? 

	updateCheckReq := &api.CatalogRegistration{
		Node:           clusterARN,
		SkipNodeUpdate: true,
		Checks:         api.HealthChecks{check},
	}
	
	clusterARN is common, skipNodeupdate is common (hardcoded) just that the checks are per service. rather than doing register for all checks. lets combine the checks and delegate to a register command in the end. 
	
@kswap

anandmukul93 (changes requested):
combined checks can we do at the end.

anandmukul93 (approved):
lgtm

PR #347, opened 2026-06-04
dependabot wf added for action everyday

Manishakumari-hc:
* Added dependabot.yml for the dependaboth to run daily and raise PRs if needed.


santoshpulluri (changes requested):
@Manishakumari-hc The dependabot workflow changes would be same everywhere.

1. Adding Changelog
2. Auto-Approval
3. Grouping multiple libraries under single PR.

PR #349, opened 2026-06-11
ci: fix coverage badge publish to use git worktree

Surabhi-1605:
## Problem

The `coverage-report` job is failing on `main` (e.g. [this run](https://github.com/hashicorp/consul-ecs/actions/runs/27295946574/job/80629474376)):

```
error: The following untracked working tree files would be overwritten by checkout:
   coverage.svg
Aborting
```

The badge step wrote `coverage.svg` into the working tree and then ran `git checkout -B badges origin/badges`. Once the `badges` branch exists and already tracks `coverage.svg`, switching to it would overwrite the untracked file git just wrote, so git aborts. This only triggers on the **second and later** pushes to `main` (the first run took the orphan-branch path), which is why it wasn't caught at merge time.

### DIFF
--- a/.github/workflows/reusable-coverage-report.yml
+++ b/.github/workflows/reusable-coverage-report.yml
@@ -158,46 +158,60 @@ jobs:
           TOTAL: ${{ steps.coverage.outputs.total }}
           GH_TOKEN: ${{ github.token }}
         run: |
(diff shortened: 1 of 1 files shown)


kswap (approved):
LGTM

PR #350, opened 2026-06-24
health-sync: batch Consul health check updates into a single atomic Catalog.Register

LordAbhishek:
**health-sync: batch Consul check updates into a single atomic Catalog.Register**
_Extension of PR: [Fix health-sync logic](https://github.com/hashicorp/consul-ecs/pull/320)_

During consul-ecs to consul healthSync, previously each tick issued up to three separate `Catalog.Register` calls (the service check, the dataplane service check, and the dataplane proxy check), which could partially fail and leave the **service-side and proxy-side dataplane checks diverged**.  
Also, updating consul for each health check is an inefficient way.

**Refactored healthSync to write healthStatusUpdate delta atomically per tick:**
1. `syncChecks` and `setChecksCritical` to stage every changed check in memory and commit the whole tick in one `Catalog.Register`, so all of a tick's updates land at a single Raft index atomically.
2. `syncChecks` now truly follows **Gather state → Compute checks (Stage) → Update Consul (Register/Commit)** flow.

**Atomic failure handling:** on a failed CatalogRegistration (write to consul), the entire batch is dropped (atomic — nothing is committed) and `currentStatuses` is left untouched, so the next tick recomputes and retries the full set. There is never a partial commit to reconcile.

**The delta guard is preserved:** only checks whose status changed since the last tick are written, and a tick with no changes performs no write. `currentStatuses` is committed only after a successful register so a failed write is retried on the next tick. For example, when only the dataplane fails, just the two dataplane checks are flipped together in one Register (sharing a single `ModifyIndex` that is ahead of the app check's `ModifyIndex`) while the unchanged app check stays out of the payload (its `ModifyIndex` is untouched) — giving both atomicity (no partial/inconsistent writes) and no redundant writes (only changed checks are sent).

**No change to health evaluation:** `computeOverallDataplaneHealth` (the whole-task aggregate that drives the dataplane/proxy checks) is untouched. Only how checks are written changed, not what status they get — the blast radius is the write path only.

Also:
- split `writeContainerHealth`/`handleHealthForDataplaneContainer`/`updateConsulHealthStatus` into `stageContainerHealthChecks` + `stageCheck` (in-memory staging) and `registerChecks` (the single write).
- log one line per changed container, only after the write succeeds.
- add unit tests for the atomic batched write, the app/dataplane failure asymmetry, and the setChecksCritical shutdown path.

### Test Setup:
Single ECS task with consul-dataplane, health-sync, mesh-init, and an nginx app container on Consul 1.22.0; consul-ecs image built from this branch's HEAD and pushed to ECR and verified the `ModifyIndex` registered service when its status toggles.

### How I've tested this PR:
1. Unit testcases (All Passed).
2. Integration test with consul using above setup.

Integration test result:

```
# RUNNING: validate_deployment
# Waits for the ECS service to stabilize and asserts test-service is registered in Consul with passing checks.
>>> validate_deployment: PASS

## RUNNING: test_service_and_proxy_registered
# Asserts both test-service and test-service-sidecar-proxy catalog entries are registered after a task starts by mesh-init.
>>> test_service_and_proxy_registered: PASS

## RUNNING: test_bug1_no_redundant_writes
# Delta guard: while health status is stable, no Catalog.Register is issued (proxy check ModifyIndex does not advance across multiple ticks).
>>> test_bug1_no_redundant_writes: PASS

## RUNNING: test_bug2_missing_container_health
# Asserts the proxy check is CRITICAL while the app container health is not yet reported (no premature passing during startup).
>>> test_bug2_missing_container_health: PASS

## RUNNING: test_app_unhealthy_marks_proxy_critical
# When the app container goes UNHEALTHY, health-sync propagates it to the proxy check (proxy goes critical), and recovers to passing once the app is healthy again.
>>> test_app_unhealthy_marks_proxy_critical: PASS

## RUNNING: test_status_change_triggers_write
# A genuine UNHEALTHY→HEALTHY transition produces exactly one Register write (ModifyIndex advances once) and
 then stays stable — the change is written, redundant writes are not.
>>> test_status_change_triggers_write: PASS

## RUNNING: test_atomic_batched_register
# Core of this PR: all checks that change in one tick are committed in a single atomic Consul Register (they share one 
ModifyIndex), unchanged ticks write nothing, and a dataplane-only failure flips just the two dataplane checks together while
the app check stays out of the payload (its ModifyIndex is untouched) — proving no partial/inconsistent writes (atomicity)
and no redundant writes (only changed checks are sent).
>>> test_atomic_batched_register: PASS

## RUNNING: test_graceful_shutdown_deregisters
# On graceful shutdown (SIGTERM), setChecksCritical marks all checks critical, then both test-service and test-service-sidecar-proxy are deregistered from the Consul catalog.
>>> test_graceful_shutdown_deregisters: PASS
```

### Consul does honor multiple checks as part of single catalog registration
It converts single health check into a slice supporting both single and multiple checks for catalog registration [See code ref](https://github.com/hashicorp/consul/blob/52d5a78d0a4f570d9f55cf454cbc01ec41ba3cd2/agent/consul/catalog_endpoint.go#L108).
Also, attached the integration test bw consul and consul-ecs logs below.

```
	// Move the old format single check into the slice, and fixup IDs.
	if args.Check != nil {
		args.Checks = append(args.Checks, args.Check)
		args.Check = nil
	}
	for _, check := range args.Checks {
		if check.Node == "" {
			check.Node = args.Node
		}
		checkPreApply(check)

		// Populate check type for cases when a check is registered in the catalog directly
		// and not via anti-entropy
		if check.Type == "" {
			chkType := check.CheckType()
			check.Type = chkType.Type()
		}
	}
```

Integration test b/w consul and consul-ecs logs:

New setup with consul health check api + service reachability check (with 2 services- test-service and client-service):
[test_atomic_batched_register_dumps3.log](https://github.com/user-attachments/files/29287738/test_atomic_batched_register_dumps3.log)
[test_atomic_batched_register.sh](https://github.com/user-attachments/files/29287739/test_atomic_batched_register.sh)

Old setup only with consul health check api:
[test_atomic_batched_register_dumps.log](https://github.com/user-attachments/files/29283071/test_atomic_batched_register_dumps.log)
[test_atomic_batched_register.sh](https://github.com/user-attachments/files/29283072/test_atomic_batched_register.sh)


subcommand/health-sync/checks.go
```diff
@@ -166,57 +214,71 @@ func computeOverallDataplaneHealth(resolvedStatuses map[string]string) string {
...
-	err := c.updateConsulHealthStatus(consulClient, checkID, clusterARN, ecsHealthStatus)
-	if err != nil {
-		return err
+	// Only the consul-dataplane container of a non-gateway service
+	// has a second (sidecar proxy) check to keep in sync.
+	if containerName != config.ConsulDataplaneContainerName || c.config.IsGateway() {
+		return api.HealthChecks{serviceCheck}, nil
 	}
 
 	proxySvcID, _ := makeProxySvcIDAndName(serviceID, "")
```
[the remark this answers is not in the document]
LordAbhishek: Not required.

subcommand/health-sync/checks.go
```diff
@@ -166,57 +214,71 @@ func computeOverallDataplaneHealth(resolvedStatuses map[string]string) string {
...
 	if !ok {
-		return fmt.Errorf("unable to find check with ID %s", checkID)
+		return nil, fmt.Errorf("unable to find check with ID %s", checkID)
 	}
 
-	check.Status = consulHealthStatus
+	check.Status = ecsHealthToConsulHealth(ecsHealthStatus)
 	check.Output = fmt.Sprintf("ECS health status is %q for container %q", ecsHealthStatus, checkID)
-	c.checks[checkID] = check
+	return check, nil
```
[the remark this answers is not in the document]
LordAbhishek: Not required.

subcommand/health-sync/checks.go
```diff
@@ -166,57 +214,71 @@ func computeOverallDataplaneHealth(resolvedStatuses map[string]string) string {
...
 	}
 
-	checkID = constructCheckID(serviceID, containerName)
-	err := c.updateConsulHealthStatus(consulClient, checkID, clusterARN, ecsHealthStatus)
-	if err != nil {
-		return err
+	// Only the consul-dataplane container of a non-gateway service
+	// has a second (sidecar proxy) check to keep in sync.
+	if containerName != config.ConsulDataplaneContainerName || c.config.IsGateway() {
+		return api.HealthChecks{serviceCheck}, nil
```
krutibaraiya: Consider simplifying the condition to read more directly:
```
if containerName == config.ConsulDataplaneContainerName && !c.config.IsGateway() {
    // add proxy check
}
return api.HealthChecks{serviceCheck}, nil
```

subcommand/health-sync/checks.go
```diff
@@ -115,31 +129,65 @@ func (c *Command) syncChecks(consulClient *api.Client,
...
+	if len(healthCheckBatch) == 0 {
+		return currentStatuses
+	}
+
+	// Commit the whole tick's staged healthCheckBatch to Consul in one atomic Catalog.Register.
+	// On failure we leave currentStatuses untouched so the next tick retries.
+	if err := c.registerChecks(consulClient, clusterARN, healthCheckBatch); err != nil {
+		c.log.Warn("failed to update Consul health status", "err", err)
+		return currentStatuses
+	}
```
krutibaraiya, line 178: Can we please ensure we have a test for this case - if `registerChecks` fails, no `currentStatuses` entry is updated.

subcommand/health-sync/checks.go
```diff
@@ -115,31 +129,65 @@ func (c *Command) syncChecks(consulClient *api.Client,
...
+		statusChangeBatch = append(statusChangeBatch, statusChange{
+			containerName:  containerName,
+			desiredStatus:  desiredStatus,
+			previousStatus: currentStatuses[containerName],
+		})
+	}
+
+	if len(healthCheckBatch) == 0 {
+		return currentStatuses
+	}
```
krutibaraiya, line 171: I think this check might be irrelevant because `registerChecks` is short-circuiting for empty slice

subcommand/health-sync/checks.go
```diff
@@ -115,31 +129,65 @@ func (c *Command) syncChecks(consulClient *api.Client,
...
+		desiredStatus  string
+		previousStatus string
+	}
+	var (
+		healthCheckBatch  api.HealthChecks // For writing (or registering) health checks (updates) to Consul
+		statusChangeBatch []statusChange   // For logging containers status change after a successful write
+	)
+	for _, containerName := range parsedContainerNames {
+		desiredStatus := resolvedStatuses[containerName]
+		if desiredStatus == currentStatuses[containerName] {
```
anandmukul93, line 148: [Blocking]
Parsedcontainernames gives two arrays 
containerstosync (which are in ecs as well as in ecs provided config) and missingcontainers(containers not in ecs but are in parsed) 
resolvedstatus only present for containerstosync. 
if resolvedstatus is not present due to missing containers in ecs metadata, i believe default vs default value check can cause it to not register essential checks. 


**READ - 
the ECS taskmetadata Containers array can be temporarily incomplete versus the task definition.
Common cases:
During startup, some containers are not created yet.
During shutdown, some containers may already be gone.
Metadata is runtime state from the ECS agent, not a strict copy of task definition.**

subcommand/health-sync/checks.go
```diff
@@ -115,31 +129,65 @@ func (c *Command) syncChecks(consulClient *api.Client,
...
 			continue
 		}
 
-		if err := c.writeContainerHealth(consulClient, taskMeta, clusterARN, name, desired); err != nil {
-			c.log.Warn("failed to update Consul health status",
-				"err", err, "container", name)
+		checks, err := c.stageContainerHealthChecks(taskMeta, containerName, desiredStatus)
```
anandmukul93, line 152: the method stageContainerHealthChecks can return nil . handle it appropriately in adding to the array.

subcommand/health-sync/checks.go
```diff
@@ -166,57 +214,71 @@ func computeOverallDataplaneHealth(resolvedStatuses map[string]string) string {
...
-// the health of consul-dataplane container. We register two checks (one for the service
-// and the other for proxy) when registering a typical service to the catalog. Updates
-// should also happen twice in such cases.
-func (c *Command) handleHealthForDataplaneContainer(consulClient *api.Client, taskID, serviceName, clusterARN, containerName, ecsHealthStatus string) error {
-	var checkID string
-	serviceID := makeServiceID(serviceName, taskID)
-	if c.config.IsGateway() {
-		checkID = constructCheckID(serviceID, containerName)
-		return c.updateConsulHealthStatus(consulClient, checkID, clusterARN, ecsHealthStatus)
+	serviceCheck, err := c.stageCheck(serviceID, containerName, ecsHealthStatus)
```
anandmukul93, line 247: first time staging for first time sync should not fail as c.checks can never be as we do 
	c.checks, err = c.fetchHealthChecks(consulClient, taskMeta)
 first thing. 

but still should handle nil checks anyways.

subcommand/health-sync/checks.go
anandmukul93: here also servicecheck  nil from c.stageCheck should be handled appropriately. 
also , the containers that mesh-init creates first time checks if you see as below - 


```
	if c.config.Gateway != nil && c.config.Gateway.Kind != "" {
		proxyRegistration = c.constructGatewayProxyRegistration(taskMeta, clusterARN)
	} else {
		serviceRegistration = c.constructServiceRegistration(taskMeta, clusterARN)
		proxyRegistration = c.constructProxyRegistration(serviceRegistration, taskMeta, clusterARN)
	}
```

so total 4 kinds of containers. 
1 - task definition healthsynccontianers (other than service if present) in ecs-provider terraform 
2 - dataplane along with service check
3 - only dataplane for sidecar proxy 
4 - gateway proxy - only dataplane. 


so either 2 dataplanes (sidecar & service )  + service or dataplane only (gateway) 
OR if other  health sync containers are present they are added as well their checks are registered for serviceId + containername.

subcommand/health-sync/checks.go
```diff
@@ -166,57 +214,71 @@ func computeOverallDataplaneHealth(resolvedStatuses map[string]string) string {
...
+		return nil, err
 	}
 
-	checkID = constructCheckID(serviceID, containerName)
-	err := c.updateConsulHealthStatus(consulClient, checkID, clusterARN, ecsHealthStatus)
-	if err != nil {
-		return err
+	// Only the consul-dataplane container of a non-gateway service
+	// has a second (sidecar proxy) check to keep in sync.
+	if containerName != config.ConsulDataplaneContainerName || c.config.IsGateway() {
```
anandmukul93: this condition is not purely clear. 
Earlier behavior as i see - 

```
serviceName := c.constructServiceName(taskMeta.Family)
	if name == config.ConsulDataplaneContainerName {
		return c.handleHealthForDataplaneContainer(consulClient, taskMeta.TaskID(), serviceName, clusterARN, name, status)
	}
	serviceID := makeServiceID(serviceName, taskMeta.TaskID())
	return c.updateConsulHealthStatus(consulClient, constructCheckID(serviceID, name), clusterARN, status)
```


if gateway - 
register a dataplane check  only 
if service - 
register a proxy dataplane check + service dataplane check for dataplane  + service check. 

New behaviour - 
if container is not dataplane and it is gateway  - this additional thought is introduced. now the understanding is not clear whether that case can also exist. condition should be positive and reduce inverted case set handling. return nil and handle at previous layer. 

There is not change in behavior but cases become underlying assumptions. 
- make it explicit - if gateway (only dataplane is registered so the container should be dataplane ) and we return check then only. 

- if it is not gateway it is service 
if container is dataplane return 2 checks for prxy and svc 

- if it is not gateway and container is service container (bool servicecontainer = !dataplanecontainer
return only service check.

subcommand/health-sync/checks.go
```diff
@@ -166,57 +214,71 @@ func computeOverallDataplaneHealth(resolvedStatuses map[string]string) string {
...
+		return api.HealthChecks{serviceCheck}, nil
 	}
 
 	proxySvcID, _ := makeProxySvcIDAndName(serviceID, "")
-	checkID = constructCheckID(proxySvcID, containerName)
-	return c.updateConsulHealthStatus(consulClient, checkID, clusterARN, ecsHealthStatus)
+	proxyCheck, err := c.stageCheck(proxySvcID, containerName, ecsHealthStatus)
```
anandmukul93: func (c *Command) fetchHealthChecks(consulClient *api.Client, taskMeta awsutil.ECSTaskMeta) (map[string]*api.HealthCheck, error) {


only fetches proxy and svc checks not dataplane checks those can be missing. handling nil is essential

subcommand/health-sync/checks.go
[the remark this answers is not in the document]
anandmukul93: as check is stored locally. what if it is not. servicename based retrieval is not supported in command.checks array. 
if not present you have to create a fresh check for dataplane for the service. 
@LordAbhishek it is important to confirm this.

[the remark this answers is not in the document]
anandmukul93: dont assume dataplane checks are present locally as i said. only proxy and svc checks are fetched in beginning.

anandmukul93:
The testing criterias are not covering observable evidence. 
Test name and pass doesnt confirm anything. 
example - 
1 - checks registered for service - get from consul catalog. 
2 -checks updated on container lifecycle - get from consul catalog. 
3 - logs from the init and health sync container for checks scenarios.

subcommand/health-sync/checks.go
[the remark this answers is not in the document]
LordAbhishek: >dont assume dataplane checks are present locally as i said. only proxy and svc checks are fetched in beginning.

No, we fetch all the checks (dataplane checks as well). So, while `serviceCheck` registration, 2 checks are registered (`serviceID-app` + `serviceID-consul-dataplane`) and for `proxyCheck` registration, a `proxySvcID-consul-dataplane` is registered. 
While fetching, we get all of them. serviceCheck return those 2 checks and proxyCheck return 1 check. Total 3 checks recieved, what was initially registered. 

Also, regarding first point: 
1. **For serviceName based retrieval is not supported**: Initially, these checks are pushed to consul by mesh-init and later pulled by healthSync container and post that it maintains local cache of those checks. So, the whole healthSync is dependent on pulling the checks from consul. 
When we register service check, we use `ServiceName` to construct the key called **checkID**, that key is used to write checks to consul from `mesh-init` container and while fetching it back, that key is reconstructed from `serviceName` in `healthSync` container and passed as filter to consul while querying. 
The **checkID** key is in form of `ServiceName-taskID-containerName` i guess, I need to check the code for exact.
So, **yes `c.checks` is keyed by CheckID, not serviceName, but that's not a limitation, because the CheckID is a pure function of (serviceName, taskID, containerName) and reconstructable.** Whenever we "want the check for a service", we reconstruct its exact CheckID from the serviceName (serviceID = serviceName-taskID, then serviceID-container / serviceID-sidecar-proxy-container) and index directly. 
It's the same key mesh-init registered under, so no serviceName-indexed search is needed. 
2. **For creating fresh dataplane check**, it's not the responsibility of healthSync to find missing healthCheck, its mesh-init gaurrentee that those check will exist once it get completed. there is retry with backoff happens in mesh-init whereby services won't get registered until it succeed, so all the checks are written as part of service CatalogRegistration payload.
_If a check were genuinely absent, `stageCheck` now returns an error and that container is skipped, rather than silently writing a wrong check._
I guess that this should not be our concern as of now.

subcommand/health-sync/checks.go
LordAbhishek: Updated

subcommand/health-sync/checks.go
LordAbhishek: Added `TestSyncChecksRegisterFailureLeavesStatusesUnchanged`

subcommand/health-sync/checks.go
LordAbhishek: Yes, it's irrelevant when we consider `registerChecks` internal behavior but it is not irrelevant in `syncChecks`. `syncCheck` should have that guard as its own logic. It also makes code more readable. 
Let's keep it. 
Also, `registerChecks` handles null values as its own internal behaviors. let's say someone wants to reuse it later.

subcommand/health-sync/checks.go
LordAbhishek: > resolvedstatus only present for containerstosync.

As per code `resolvedstatus` contains all containers. For every `containersToSync`/foundContainers, it assigns their ECS health status and for the `missingContainers`, it assigns UNHEALTHY.

> default vs default value check can cause it to not register essential checks._ 

**This never happens** because a missing container is UNHEALTHY and on the first tick `[currentStatuses[name]] == ""`, so `UNHEALTHY != ""` ⇒ it is written/marked (critical).
Eg:
Startup (container not created yet) ⇒ missing ⇒ UNHEALTHY ⇒ written critical.
Shutdown (container already gone) ⇒ missing ⇒ UNHEALTHY ⇒ written critical.
**So, a container that is missing or has unknown status can only ever resolve to CRITICAL, never to "" or false PASSING. Also, by default mesh-init registeres every check as `critical.`** 

Code Ref:
```
    resolvedStatuses := make(map[string]string, len(parsedContainerNames))
    foundContainers := make(map[string]awsutil.ECSTaskMetaContainer, len(containersToSync))
    for _, container := range containersToSync {
        resolvedStatuses[container.Name] = container.Health.Status
        foundContainers[container.Name] = container
    }
    for _, name := range missingContainers {
        resolvedStatuses[name] = string(types.HealthStatusUnhealthy)
    }
```

subcommand/health-sync/checks.go
LordAbhishek: `stageContainerHealthChecks` already returns checks in an array and an error. 
```
type HealthChecks []*HealthCheck
var checks api.HealthChecks
```

The `nil` possibility was in `stageCheck`, where the check pointer was dereferenced (check.Status / check.Output). _Now a **missing or nil check is returned as an error from stageCheck**, and every caller acts on that error — continue in syncChecks, return nil, err in `stageContainerHealthChecks`_ — so a nil check is never appended to the batch. 
Because `stageCheck` returns a non-nil error exactly when the check is missing/nil, past the `err != nil` check `serviceCheck/proxyCheck` are guaranteed non-nil, making an explicit nil check here unreachable. 
Added `TestStageCheckReturnsErrorWhenCheckMissingOrNil` unit test is added as well for this.

```
func (c *Command) stageCheck(serviceID, containerName, ecsHealthStatus string) (*api.HealthCheck, error) {
    checkID := constructCheckID(serviceID, containerName)
    check, ok := c.checks[checkID]
    if !ok || check == nil {       // <- here //
        return nil, fmt.Errorf("unable to find check with ID %s", checkID)
    }

    check.Status = ecsHealthToConsulHealth(ecsHealthStatus)
    check.Output = fmt.Sprintf("ECS health status is %q for container %q", ecsHealthStatus, checkID)
    return check, nil
}
```

subcommand/health-sync/checks.go
LordAbhishek: The `nil` possibility was in `stageCheck`, where the check pointer was dereferenced (check.Status / check.Output). _Now a **missing or nil check is returned as an error from stageCheck**, and every caller acts on that error — continue in syncChecks, return nil, err in `stageContainerHealthChecks`_ — so a nil check is never appended to the batch. Because `stageCheck` returns a non-nil error exactly when the check is missing/nil, past the `err != nil` check `serviceCheck/proxyCheck` are guaranteed non-nil, making an explicit nil check here unreachable. 
Added `TestStageCheckReturnsErrorWhenCheckMissingOrNil` unit test is added as well for this.

subcommand/health-sync/checks.go
LordAbhishek: Updated the method to explicitly separate the cases. 
And removed below behavior.
> if container is not dataplane and it is gateway - this additional thought is introduced. now the understanding is not clear whether that case can also exist.


**Health registration behavior (agreed as above):**

> if gateway - 
register a dataplane check only 
if service - 
register a proxy dataplane check + [service dataplane check for dataplane + service check].

**What i have understood so far during healthsync (container wise healthchecks):** 
```
containerName               serviceRegistrationCheck       proxyRegistrationCheck
------------------------------------------------------------------------
app                         serviceID-app                  — (skipped)
consul-dataplane (normal)   serviceID-consul-dataplane     proxySvcID-consul-dataplane
consul-dataplane (gateway)  gwServiceID-consul-dataplane   — (skipped)
```
So, 
1. when we have `app` container, we will simply return serviceCheck (`serviceID-app`)
3. when we have `dataplane` container, and its not gateway, we return both registered serviceCheck (`serviceID-app`) and proxyCheck ( `serviceID-consul-dataplane` + `proxySvcID-consul-dataplane`) both.
4. when we have `dataplane` container, and it's gateway, since dataplane do not have proxy, no proxyChecks are sent only serviceCheck(`gwServiceID-consul-dataplane`). 

**Also, `serviceCheck` (`serviceID-app` or `serviceID-consul-dataplane` or `gwServiceID-consul-dataplane` ) is keyed with `serviceID` and `serviceID` is conditionally generated for both normal service and gateway, both during regitstration and health sync.** 
Therefore, we can keep a single `serviceCheck` for all type of containers `serviceName := c.constructServiceName(taskMeta.Family)`.

subcommand/health-sync/checks.go
LordAbhishek: **`fetchHealthChecks` picks up the dataplane check**: the dataplane service readiness check is registered under the service registration, so the first query returns the app checks plus the service dataplane check, and the second query returns the proxy dataplane check — so all three land in `c.checks`. 

The `nil` possibility was in `stageCheck`, where the check pointer was dereferenced (check.Status / check.Output). _Now a **missing or nil check is returned as an error from stageCheck**, and every caller acts on that error — continue in syncChecks, return nil, err in `stageContainerHealthChecks`_ — so a nil check is never appended to the batch. Because `stageCheck` returns a non-nil error exactly when the check is missing/nil, past the `err != nil` check `serviceCheck/proxyCheck` are guaranteed non-nil, making an explicit nil check here unreachable. 
Added `TestStageCheckReturnsErrorWhenCheckMissingOrNil` unit test is added as well for this.

LordAbhishek:
> The testing criterias are not covering observable evidence. Test name and pass doesnt confirm anything. example - 1 - checks registered for service - get from consul catalog. 2 -checks updated on container lifecycle - get from consul catalog. 3 - logs from the init and health sync container for checks scenarios.

All the test was asserted by fetching consul catalog previously. I have added log line of health-sync container for verification as well.
I am not aware about the Lifecycle changes and testcases but they are also covered via consul catalog assertions.

subcommand/health-sync/checks.go
krutibaraiya: The intention is to put gateway/service split to be the outer condition, making each case self-contained. @abhishek-hashicorp I completely agree with your explanation but the conditions are nested.
Shouldn't it be

```
// Explicit separation: service type is the outer branch
if c.config.IsGateway() {
    // Gateway only ever has one check: the dataplane check
    return api.HealthChecks{serviceCheck}, nil
}

// Non-gateway (service): dataplane maps to both service and proxy checks
if containerName == config.ConsulDataplaneContainerName {
    proxySvcID, _ := makeProxySvcIDAndName(serviceID, serviceName)
    proxyCheck, err := c.stageCheck(proxySvcID, containerName, ecsHealthStatus)
    if err != nil {
        return nil, err
    }
    return api.HealthChecks{serviceCheck, proxyCheck}, nil
}

// Non-gateway, non-dataplane: app container maps to one service check
return api.HealthChecks{serviceCheck}, nil
```

subcommand/health-sync/checks.go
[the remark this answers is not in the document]
krutibaraiya: Minor: I think its still better to pass `serviceName` due to
1. Consistency with the rest of the codebase
2. Passing an empty string here feels a bit risky, as a future refactor might inadvertently start relying on the second return value

PR #353, opened 2026-07-02
mesh-init: add consul-dataplane and consul-ecs binaries version into ServiceMeta during service registration

### LINKED ISSUE #339: Inject consul-dataplane and consul-ecs binaries version into `ServiceMetadata`
#### Description

During service registration inject the current consul-dataplane and consul-ecs binaries version into `ServiceMetadada`

#### Use Cases

During LTS upgrades, sometimes it's needed to update the data-plane to a version supported by current and the next LTS version. Before actually upgrade the control-plane we need to ensure all existing mesh services are using the expected data-plane version, currently this information is not available via Consul API. 

```
curl http://localhost:8500/v1/catalog/service/foobar-sidecar-proxy | jq
...
  "ServiceMeta": {
      "source": "consul-ecs",
      "task-arn": "arn:aws:ecs:us-east-1:1111111:task/<cluster>/<task-id>",
      "task-id": "<task-id>"
    },
...
```

#### Alternative Solutions
The workaround is to describe the latest task definitions for all services and search for the data-plane version defined in container definitions.

LordAbhishek:
## Inject `consul-ecs` and `consul-dataplane` versions into ServiceMeta
Fixes #339

### 1. What is changed/added in this PR

This PR extends the Consul `ServiceMeta` that `mesh-init` sets during service, sidecar-proxy, and gateway registration with two new fields, so operators can query the exact binary versions running in a task (e.g. to verify dataplane versions before an LTS upgrade):

- `ecs-service-version` — the running `consul-ecs` binary version (`version.GetHumanVersion()`).
- `dataplane-version` — the `consul-dataplane` version, derived from its container image in the ECS Task. 

**Note on consul dataplane version:** 
1. Consul `dataplane-version` can be raw container image reference in case if image lacks tags as fallback and `dataplane-version` field will be skipped if there is no container image reference in ECS Task.
2. Consul dataplane image is read via **ECS DescribeTasks**, since the dataplane container is not yet present in the local task metadata endpoint when mesh-init registers the service. This requires the `ecs:DescribeTasks` permission on the ECS task role; when it is missing, the version is omitted (with a log) and service registration is never blocked.
3. Customer is OK to provide extra IAM permissions ([ref](https://github.com/hashicorp/consul-ecs/issues/339#issuecomment-4954369535)) and want real-time container `dataplane-version` instead of terraform-module propagated values. 

Before:
```json
"ServiceMeta": {
  "source": "consul-ecs",
  "task-arn": "arn:aws:ecs:...:task/<cluster>/<task-id>",
  "task-id": "<task-id>"
}
```
After:
```json
"ServiceMeta": {
  "source": "consul-ecs",
  "task-arn": "arn:aws:ecs:...:task/<cluster>/<task-id>",
  "task-id": "<task-id>",
  "ecs-service-version": "v0.9.4-dev (<sha>)",
  "dataplane-version": "1.3.0"
}
```

Implementation notes (no existing behavior changed):
- **No change to the existing base meta flow.** `task-id` / `task-arn` / `source` are built and merged with user config meta exactly as on `main`. **The version fields are simply applied *after* that merge via a new `setVersionMeta` helper.**
- **awsutil**: added an `Image` field to `ECSTaskContainer` and an `ImageVersion()` helper that returns the image **tag** when present, and otherwise the raw image reference (bare repo or digest-pinned), so the value is always meaningful.
- **Safety guards** (both protect against `mesh-init`'s infinite registration retry silently hanging):
  - Version meta keys deliberately avoid the `consul-` prefix, which Consul reserves for internal use and rejects during registration.
  - A value longer than Consul's 512-char meta-value limit is omitted rather than truncated (a truncated image ref would be misleading, and a full one would fail registration).

### 2. How I've tested this PR

- **Unit tests** for the new logic:
  - `TestECSTaskMetaContainerImageVersion` — covers tags, `latest`, digest-only, `host:port` registries, tag+digest, and empty image.
  - `TestGetDataplaneVersion` — including the `>512` char omission case.
  - `TestSetVersionMetaOverridesUserMeta` — verifies the version keys are not user-overridable.
- **Integration tests against a real Consul agent:** `TestRun` and `TestGateway` spin up an actual `consul` server (via `hashicorp/consul/sdk/testutil`), run the full `mesh-init` flow, register the service/proxy/gateway through the real Consul catalog, then read the records back and assert `ServiceMeta` contains the new `ecs-service-version` / `dataplane-version` values. This exercises Consul's own server-side meta validation end-to-end.
- Consul-ecs deployment test:
Adding couple of test setups scripts and logs - All race test (25 continuous deployments) passed.
[test_serviceMetadata_race.sh](https://github.com/user-attachments/files/30125496/test_serviceMetadata_race.sh)
[test_serviceMetadata_race.log](https://github.com/user-attachments/files/30125498/test_serviceMetadata_race.log)
[test_serviceMetadata.sh](https://github.com/user-attachments/files/30125497/test_serviceMetadata.sh)
[test_serviceMetadata3.log](https://github.com/user-attachments/files/30125500/test_serviceMetadata3.log)

### 3. How I expect reviewers to test this PR
- **Version fields are intentionally not user-overridable.** The `ecs-service-version` and `dataplane-version` values are owned by `consul-ecs` and applied *after* the user-provided `service.meta` / `gateway.meta`, so a user cannot override them via config. **This is deliberate: the whole purpose of these fields is to report the version **actually in use** (e.g. for upgrade audits), so allowing an override would let the reported value drift from reality.** _That said, this is a design choice — if we'd prefer to let users override them, it's a trivial change (apply the version meta *before* the user meta instead of after). Please let me know if we need to go this path._
- **Dataplane version can be raw container image reference in case if image lacks tags as fallback:** Please let me know if this needs to be changed as well. Should we only return image tags?
---

AWS ECS Ref:
https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_DescribeTasks.html
https://docs.aws.amazon.com/AmazonECS/latest/developerguide/security_iam_id-based-policy-examples.html#IAM_run_policies

### DIFF
--- a/awsutil/awsutil.go
+++ b/awsutil/awsutil.go
@@ -41,15 +41,13 @@ type ECSTaskMeta struct {
 	Cluster          string                 `json:"Cluster"`
 	TaskARN          string                 `json:"TaskARN"`
 	Family           string                 `json:"Family"`
-	Revision         string                 `json:"Revision"`
 	Containers       []ECSTaskMetaContainer `json:"Containers"`
 	AvailabilityZone string                 `json:"AvailabilityZone"`
 }
 
-// ECSTaskDefinitionAPI is the subset of the ECS API used to read a task
-// definition. It is satisfied by *ecs.Client.
-type ECSTaskDefinitionAPI interface {
-	DescribeTaskDefinition(ctx context.Context, params *ecs.DescribeTaskDefinitionInput, optFns ...func(*ecs.Options)) (*ecs.DescribeTaskDefinitionOutput, error)
+// ECSTaskAPI is the subset of the ECS API needed by mesh-init.
+type ECSTaskAPI interface {
+	DescribeTasks(ctx context.Context, params *ecs.DescribeTasksInput, optFns ...func(*ecs.Options)) (*ecs.DescribeTasksOutput, error)
 }
 
 type ECSTaskMetaContainer struct {
@@ -74,16 +72,6 @@ type ECSTaskMetaNetwork struct {
 
 func (e ECSTaskMeta) TaskID() string {
 	return ParseTaskID(e.TaskARN)
-}
-
-// TaskDefinitionID returns the task definition identifier ("family:revision")
-// for use with the ECS DescribeTaskDefinition API. It falls back to just the
-// family when the revision is unavailable.
-func (e ECSTaskMeta) TaskDefinitionID() string {
-	if e.Revision != "" {
-		return fmt.Sprintf("%s:%s", e.Family, e.Revision)
-	}
-	return e.Family
 }
 
 // ImageVersion returns a version identifier derived from a container image
@@ -115,27 +103,23 @@ func ImageVersion(image string) string {
 	return image
 }
 
-// ContainerImage returns the image reference declared for the named container
-// in the given task definition (accepts "family", "family:revision", or a full
-// task definition ARN).
-// It reads from the task definition rather than the task
-// metadata endpoint because the task definition always contains the complete,
-// static container list, regardless of container startup ordering at runtime
-// (the metadata endpoint can omit containers that have not started yet).
-// It returns "" (with no error) when the container is not present.
-func ContainerImage(ctx context.Context, client ECSTaskDefinitionAPI, taskDefinition, containerName string) (string, error) {
-	out, err := client.DescribeTaskDefinition(ctx, &ecs.DescribeTaskDefinitionInput{
-		TaskDefinition: aws.String(taskDefinition),
+// ContainerImage returns the image reference for the named container from
+// ECS DescribeTasks runtime state. It returns "" (with no error) when the
+// container is not present in the task response.
+func ContainerImage(ctx context.Context, client ECSTaskAPI, cluster, taskARN, containerName string) (string, error) {
+	out, err := client.DescribeTasks(ctx, &ecs.DescribeTasksInput{
+		Cluster: aws.String(cluster),
+		Tasks:   []string{taskARN},
 	})
 	if err != nil {
-		return "", fmt.Errorf("describing task definition %q: %w", taskDefinition, err)
+		return "", fmt.Errorf("describing task %q: %w", taskARN, err)
 	}
-	if out == nil || out.TaskDefinition == nil {
-		return "", fmt.Errorf("task definition %q not found", taskDefinition)
+	if out == nil || len(out.Tasks) == 0 {
+		return "", fmt.Errorf("task %q not found", taskARN)
 	}
-	for _, cd := range out.TaskDefinition.ContainerDefinitions {
-		if aws.ToString(cd.Name) == containerName {
-			return aws.ToString(cd.Image), nil
+	for _, container := range out.Tasks[0].Containers {
+		if aws.ToString(container.Name) == containerName {
+			return aws.ToString(container.Image), nil
 		}
 	}
 	return "", nil
--- a/awsutil/awsutil.go
+++ b/awsutil/awsutil.go
@@ -15,6 +15,7 @@ import (
 	"github.com/aws/aws-sdk-go-v2/aws"
 	"github.com/aws/aws-sdk-go-v2/aws/arn"
 	"github.com/aws/aws-sdk-go-v2/config"
+	"github.com/aws/aws-sdk-go-v2/service/ecs"
 	"github.com/aws/aws-sdk-go-v2/service/ecs/types"
 	smithymiddleware "github.com/aws/smithy-go/middleware"
 	smithyhttp "github.com/aws/smithy-go/transport/http"
@@ -40,13 +41,19 @@ type ECSTaskMeta struct {
 	Cluster          string                 `json:"Cluster"`
 	TaskARN          string                 `json:"TaskARN"`
 	Family           string                 `json:"Family"`
+	Revision         string                 `json:"Revision"`
 	Containers       []ECSTaskMetaContainer `json:"Containers"`
 	AvailabilityZone string                 `json:"AvailabilityZone"`
 }
 
+// ECSTaskDefinitionAPI is the subset of the ECS API used to read a task
+// definition. It is satisfied by *ecs.Client.
+type ECSTaskDefinitionAPI interface {
+	DescribeTaskDefinition(ctx context.Context, params *ecs.DescribeTaskDefinitionInput, optFns ...func(*ecs.Options)) (*ecs.DescribeTaskDefinitionOutput, error)
+}
+
 type ECSTaskMetaContainer struct {
 	Name          string               `json:"Name"`
-	Image         string               `json:"Image"`
 	Health        ECSTaskMetaHealth    `json:"Health"`
 	DesiredStatus string               `json:"DesiredStatus"`
 	KnownStatus   string               `json:"KnownStatus"`
@@ -69,16 +76,25 @@ func (e ECSTaskMeta) TaskID() string {
 	return ParseTaskID(e.TaskARN)
 }
 
-// ImageVersion returns a version identifier derived from the
-// container's image reference. It prefers the image tag (e.g. "1.3.0").
-// When the image has no tag (it is pinned only by digest, or references a bare
-// repository), it returns the raw image reference unchanged, so the value is
-// still meaningful and can be used to pull the exact image.
-// It returns "" only when no image reference is available.
-func (c ECSTaskMetaContainer) ImageVersion() string {
-	imageRef := c.Image      // full reference, e.g. "localhost:5000/repo:1.3.0@sha256:..."
-	imageWithTag := imageRef // reference reduced to its final "repo:tag" segment
-	tag := ""                // the tag portion, if the image has one
+// TaskDefinitionID returns the task definition identifier ("family:revision")
+// for use with the ECS DescribeTaskDefinition API. It falls back to just the
+// family when the revision is unavailable.
+func (e ECSTaskMeta) TaskDefinitionID() string {
+	if e.Revision != "" {
+		return fmt.Sprintf("%s:%s", e.Family, e.Revision)
+	}
+	return e.Family
+}
+
+// ImageVersion returns a version identifier derived from a container image
+// reference. It prefers the image tag (e.g. "1.3.0"). When the image has no tag
+// (it is pinned only by digest, or references a bare repository), it returns
+// the raw image reference unchanged, so the value is still meaningful and can
+// be used to pull the exact image.
+// It returns "" only when image is empty.
+func ImageVersion(image string) string {
+	imageWithTag := image // reference reduced to its final "repo:tag" segment
+	tag := ""             // the tag portion, if the image has one
 
 	// Drop any digest ("@sha256:...") so it isn't mistaken for a tag.
 	if idx := strings.Index(imageWithTag, "@"); idx != -1 {
@@ -96,7 +112,33 @@ func (c ECSTaskMetaContainer) ImageVersion() string {
 		return tag
 	}
 	// No tag: return the raw reference (bare repo or digest-pinned) as-is.
-	return imageRef
+	return image
+}
+
+// ContainerImage returns the image reference declared for the named container
+// in the given task definition (accepts "family", "family:revision", or a full
+// task definition ARN).
+// It reads from the task definition rather than the task
+// metadata endpoint because the task definition always contains the complete,
+// static container list, regardless of container startup ordering at runtime
+// (the metadata endpoint can omit containers that have not started yet).
+// It returns "" (with no error) when the container is not present.
+func ContainerImage(ctx context.Context, client ECSTaskDefinitionAPI, taskDefinition, containerName string) (string, error) {
+	out, err := client.DescribeTaskDefinition(ctx, &ecs.DescribeTaskDefinitionInput{
+		TaskDefinition: aws.String(taskDefinition),
+	})
+	if err != nil {
+		return "", fmt.Errorf("describing task definition %q: %w", taskDefinition, err)
+	}
+	if out == nil || out.TaskDefinition == nil {
+		return "", fmt.Errorf("task definition %q not found", taskDefinition)
+	}
+	for _, cd := range out.TaskDefinition.ContainerDefinitions {
+		if aws.ToString(cd.Name) == containerName {
+			return aws.ToString(cd.Image), nil
+		}
+	}
+	return "", nil
 }
 
 func (e ECSTaskMeta) ClusterARN() (string, error) {
--- a/awsutil/awsutil.go
+++ b/awsutil/awsutil.go
@@ -112,7 +112,7 @@ func ContainerImage(ctx context.Context, client ECSTaskAPI, cluster, taskARN, co
 		Tasks:   []string{taskARN},
 	})
 	if err != nil {
-		return "", fmt.Errorf("describing task %q: %w", taskARN, err)
+		return "", fmt.Errorf("describing ECS task %q: %w", taskARN, err)
 	}
 	if out == nil || len(out.Tasks) == 0 {
 		return "", fmt.Errorf("task %q not found", taskARN)
--- a/awsutil/awsutil.go
+++ b/awsutil/awsutil.go
@@ -46,6 +46,7 @@ type ECSTaskMeta struct {
 
 type ECSTaskMetaContainer struct {
 	Name          string               `json:"Name"`
+	Image         string               `json:"Image"`
 	Health        ECSTaskMetaHealth    `json:"Health"`
 	DesiredStatus string               `json:"DesiredStatus"`
 	KnownStatus   string               `json:"KnownStatus"`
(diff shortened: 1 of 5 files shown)


subcommand/mesh-init/command.go
```diff
@@ -539,6 +572,48 @@ func getNodeMeta() map[string]string {
...
+
+// getDataplaneVersion returns the version of the consul-dataplane container,
+// derived from its image reference in ECS DescribeTasks runtime state.
+//
+// This is best-effort and never blocks registration.
+// It returns "" when the container is absent or the derived value exceeds
+// Consul's meta value limit, and additionally logs a warning when the task
+// lookup itself fails.
+func (c *Command) getDataplaneVersion(taskMeta awsutil.ECSTaskMeta) string {
+	dpImage, err := awsutil.ContainerImage(context.Background(), c.ecsClient, taskMeta.Cluster, taskMeta.TaskARN, config.ConsulDataplaneContainerName)
```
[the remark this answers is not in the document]
krutibaraiya: @LordAbhishek Please check this

subcommand/mesh-init/command.go
```diff
@@ -85,6 +101,15 @@ func (c *Command) realRun() error {
...
 	}
 
+	// Set up the ECS client used to read container image data from DescribeTasks.
+	if c.ecsClient == nil {
+		awsCfg, err := awsutil.NewAWSConfig(taskMeta, "mesh-init")
+		if err != nil {
+			return fmt.Errorf("constructing aws config: %w", err)
+		}
+		c.ecsClient = ecs.NewFromConfig(awsCfg)
+	}
```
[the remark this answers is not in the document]
krutibaraiya: @LordAbhishek This one also

krutibaraiya:
Please add one line changelog in CHANGELOG.md under "Unreleased" section

subcommand/mesh-init/command.go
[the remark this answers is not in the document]
LordAbhishek: Fixed it.

subcommand/mesh-init/command.go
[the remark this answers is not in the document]
LordAbhishek: Fixed it as well.

subcommand/mesh-init/command.go
```diff
@@ -539,6 +572,48 @@ func getNodeMeta() map[string]string {
...
+//
+// Meta keys must not start with "consul-": that prefix is reserved for Consul's
+// internal use and Consul rejects such keys during service registration.
+// Refer: https://github.com/hashicorp/consul/blob/9a38fac228fae7960f12f5b2a45c7548c90e8224/agent/structs/structs.go#L182
+func (c *Command) setVersionMeta(meta map[string]string, taskMeta awsutil.ECSTaskMeta) {
+	meta["ecs-service-version"] = version.GetHumanVersion()
+	if dp := c.getDataplaneVersion(taskMeta); dp != "" {
+		meta["dataplane-version"] = dp
+	}
+}
```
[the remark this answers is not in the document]
LordAbhishek: Fixed it as well.

subcommand/mesh-init/command.go
```diff
@@ -539,6 +581,66 @@ func getNodeMeta() map[string]string {
...
+// internal use and Consul rejects such keys during service registration.
+// Refer: https://github.com/hashicorp/consul/blob/9a38fac228fae7960f12f5b2a45c7548c90e8224/agent/structs/structs.go#L182
+func (c *Command) setVersionMeta(meta map[string]string, taskMeta awsutil.ECSTaskMeta) {
+	if ecsVersion := version.GetHumanVersion(); len(ecsVersion) <= maxMetaValueLength {
+		meta["ecs-service-version"] = ecsVersion
+	}
+	if dp := c.getDataplaneVersion(taskMeta); dp != "" {
+		meta["dataplane-version"] = dp
+	}
+}
```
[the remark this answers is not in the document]
LordAbhishek: Fixed it as well.

awsutil/awsutil.go
```diff
@@ -68,6 +74,57 @@ func (e ECSTaskMeta) TaskID() string {
...
+	if err != nil {
+		return "", fmt.Errorf("describing task %q: %w", taskARN, err)
+	}
+	if out == nil || len(out.Tasks) == 0 {
+		return "", fmt.Errorf("task %q not found", taskARN)
+	}
+	for _, container := range out.Tasks[0].Containers {
+		if aws.ToString(container.Name) == containerName {
+			return aws.ToString(container.Image), nil
+		}
```
krutibaraiya, line 123: Can we add some tests covering these cases? Currently I see that it is only called indirectly in the tests via `fakeECSClient`, but it would be good to cover all these conditions using a table-driven test with a fake `ECSTaskAPI`.
1. task with mulitple containers where DP isn't first -> loop/skip logic
2. matching container present with empty image -> return nil 
3. If output is nil or length of tasks is 0 -> return "task %q not found" error

awsutil/awsutil.go
```diff
@@ -44,6 +45,11 @@ type ECSTaskMeta struct {
 	AvailabilityZone string                 `json:"AvailabilityZone"`
 }
 
+// ECSTaskAPI is the subset of the ECS API needed by mesh-init.
+type ECSTaskAPI interface {
+	DescribeTasks(ctx context.Context, params *ecs.DescribeTasksInput, optFns ...func(*ecs.Options)) (*ecs.DescribeTasksOutput, error)
```
krutibaraiya, line 50: Can you verify on real task/e2e setup that this returns dataplane image at mesh-init time? According to my understanding, dataplane container depends on mesh-init `SUCCESS`, so mesh-init might run and call `DescribeTasks` before dataplane is created. It might be possible that DP container may not be in the runtime containers and have an empty image-> leading to silent omission.

Ignore if already verified

LordAbhishek: Hi @krutibaraiya  
```
$ aws ecs describe-task-definition --region ap-south-1 --task-definition test-service --query 'taskDefinition.containerDefinitions[].{name:name,dependsOn:dependsOn}' --output json | cat
[
    {
        "name": "app",
        "dependsOn": [
            {
                "containerName": "consul-dataplane",
                "condition": "HEALTHY"
            }
        ]
    },
    {
        "name": "consul-ecs-mesh-init",
        "dependsOn": null
    },
    {
        "name": "consul-dataplane",
        "dependsOn": [
            {
                "containerName": "consul-ecs-mesh-init",
                "condition": "SUCCESS"
            }
        ]
    },
    {
        "name": "consul-ecs-health-sync",
        "dependsOn": [
            {
                "containerName": "consul-ecs-mesh-init",
                "condition": "SUCCESS"
            }
        ]
    }
]
```
As per task definition, dataplane won't be created until mesh-init finishes. The DP container stays in Pending state until mesh-init finishes. I am not exactly sure, but ECS builds that runtime list as it provisions the tasks (from `ecs:DescribeTaskDefinition`), copying each container's static image into it. So, ideally, an empty image for a listed container basically never happens. 
Other race condition can be mesh-init starts before ECS starts provisioning the task, this is also ruled out bcoz mesh-init itself is result of its ECS task provisioning. 
Also, we have 5sec timeout for AWS SDKs when polling ecs task. That time should be enough to handle any delays. 

Adding couple of test setups scripts and logs - All race test (25 continuous deployments) passed.
[test_serviceMetadata_race.sh](https://github.com/user-attachments/files/30125496/test_serviceMetadata_race.sh)
[test_serviceMetadata_race.log](https://github.com/user-attachments/files/30125498/test_serviceMetadata_race.log)
[test_serviceMetadata.sh](https://github.com/user-attachments/files/30125497/test_serviceMetadata.sh)
[test_serviceMetadata3.log](https://github.com/user-attachments/files/30125500/test_serviceMetadata3.log)

awsutil/awsutil.go
LordAbhishek: Added

awsutil/awsutil.go
anandmukul93: It would make more sense if the dataplane container availability here can be logged tomorrow. so that we have the understanding from log itself. otherwise, my understanding says. 
https://github.com/hashicorp/terraform-aws-consul-ecs/blob/da2db1edcd7286c0ecf5436a4f1ea2ea2e6d2171/modules/mesh-task/main.tf#L223

All updates are assumed to be done by terraform apply . then why do we even this flow ? 
are we saying that people will use aws ui to update task definitions and run ? if so ,the full auditability will go for a toss. 

also, shed some light on if dataplane has a dependency on the mesh-init then mesh-init will only pick it up from task definition which is static and applied by terraform apply in ecs. right?

subcommand/mesh-init/command.go
```diff
@@ -260,6 +294,10 @@ func (c *Command) constructServiceRegistration(taskMeta awsutil.ECSTaskMeta, clu
 		"source":   "consul-ecs",
 	}, c.config.Service.Meta)
```
anandmukul93, line 295: service.meta can contain the details injected in ecs tf module itself. are we targetting ui based changes to task definition to be propagated here. that breaks auditability .

subcommand/mesh-init/command.go
```diff
@@ -539,6 +581,72 @@ func getNodeMeta() map[string]string {
...
+// These reflect the actual running binaries, so they are applied after the
+// user-provided meta and cannot be overridden: the purpose of these fields is
+// to report the exact versions in use (e.g. for LTS upgrade audits).
+// The dataplane-version key is omitted when the version cannot be determined,
+// so we never advertise a misleading empty value.
+//
+// Meta keys must not start with "consul-": that prefix is reserved for Consul's
+// internal use and Consul rejects such keys during service registration.
+// Refer: https://github.com/hashicorp/consul/blob/9a38fac228fae7960f12f5b2a45c7548c90e8224/agent/structs/structs.go#L182
+func (c *Command) setVersionMeta(meta map[string]string, taskMeta awsutil.ECSTaskMeta) {
```
anandmukul93, line 594: @LordAbhishek @rashmirathi-2 @krutibaraiya 
my thought is all of this is not required if we say that customer is using the terraform modify and apply path. 
```mermaid
%%{init: { 'theme': 'base', 'flowchart': { 'htmlLabels': true } }%%
graph TD
    classDef default fill:#f9f9f9,stroke:#333,stroke-width:1px,width:220px;

    TF{{"Terraform module
consul_ecs_image = '...consul-ecs:0.10.0'
consul_dataplane_image = '...consul-dataplane:2.0.1'"}} -->|"encodes CONSUL_ECS_CONFIG_JSON env var"| TD{{"ECS Task Definition
(service.meta = consul_service_meta)"}}

    TD -->|"container starts"| MeshInit{{"mesh-init runs
(constructServiceRegistration)"}}

    MeshInit -->|"reads c.config.Service.Meta"| UserMeta{{"user-supplied meta"}}

    MeshInit -->|"setVersionMeta calls getDataplaneVersion"| ECS_API{{"DescribeTasks API call
→ reads container.Image at runtime"}}

    UserMeta --> Consul{{"Consul service registration
ecs-service-version + dataplane-version"}}
    
    ECS_API --> Consul
```
the config is passed via env variable set action to the task definition. 
https://github.com/hashicorp/terraform-aws-consul-ecs/blob/da2db1edcd7286c0ecf5436a4f1ea2ea2e6d2171/modules/mesh-task/config.tf#L6

if you can merge variablees for dp version and ecs version here. then thats best. no changes related to code are required here. 

you have to add validations on the variables for what are additionally allowed. 

https://github.com/hashicorp/terraform-aws-consul-ecs/blob/da2db1edcd7286c0ecf5436a4f1ea2ea2e6d2171/modules/mesh-task/variables.tf#L378-L386

see above example.

awsutil/awsutil.go
v-rosa: @anandmukul93 

Given I raised this GH issue, let me clarify our use case.

> All updates are assumed to be done by terraform apply . then why do we even this flow ?
> are we saying that people will use aws ui to update task definitions and run ? if so ,the full auditability will go for a toss.

Our deployment model, relies on a **template** task-definition per service using your [terraform-aws-consul-ecs module](https://github.com/hashicorp/terraform-aws-consul-ecs/). Then the GH deployment workflow picks the latest service task-def template and allows some overrides during deploy time, e.g workload container version, env-vars, cpu, memory, ...,  and the consul side cars versions, usually this flow is useful when we want to upgrade minor versions of dataplane, we use this approach to keep things up-to-date with minimal effort.

This override mechanism is a well known pattern in the ECS community to decouple the infrastructure configuration (which includes the task-def) with the actual workload container version. I can elaborate a bit more on this if necessary.

So, not always the terraform task-def created by the terraform-aws-consul-ecs is the complete source of truth.

I do think, describing the running task definition covers both use cases:
- terraform-aws-consul-ecs changes only
- override task-def pattern.

subcommand/mesh-init/command.go
v-rosa: @anandmukul93 please check the conversation here: https://github.com/hashicorp/consul-ecs/issues/339

and my clarification here: https://github.com/hashicorp/consul-ecs/pull/353#discussion_r3764746284

awsutil/awsutil.go
anandmukul93: @v-rosa 
i may not have a full context here. do you mean this by the task-def override ? 
`
[Phase 1: Infra Setup]
  Terraform runs 
    └─ Creates Baseline Task Definition (v1)
    └─ Creates ECS Service (Pointed to v1, ignoring future version updates)

[Phase 2: App Deployment]
  GitHub Actions Workflow runs
    └─ Fetches current baseline configuration 
    └─ Overrides app version, env-vars, cpu, memory, sidecar versions
    └─ Registers a NEW standalone Task Definition via AWS CLI/API (v2)
    └─ Updates ECS Service to point to v2 via AWS CLI/API
    `
even in that isn't adding meta in overrides a simpler way given that it ecs binary create any logical understanding other than to pass it to consul service definition which is the main requirement as i believed. in the end the config in the task definition that is overriden will be picked by ecs in runtime through task definition anyway. i am not totally sure whether this amount of change is necessary. 
@LordAbhishek

LordAbhishek: Hi @v-rosa and @anandmukul93, 

> Alternative Solutions
The workaround is to describe the latest task definitions for all services and search for the data-plane version defined in container definitions.

We have an exact `meta` config in terraform-module [terraform-module-config](
https://github.com/hashicorp/terraform-aws-consul-ecs/blob/da2db1edcd7286c0ecf5436a4f1ea2ea2e6d2171/modules/mesh-task/config.tf#L49) which accepts `key and value` [terraform-module-variables](
https://github.com/hashicorp/terraform-aws-consul-ecs/blob/da2db1edcd7286c0ecf5436a4f1ea2ea2e6d2171/modules/mesh-task/variables.tf#L26) and that is directly [copied & pasted](https://github.com/hashicorp/consul-ecs/blob/a371d860fbf575a01c7057082660f5992f126637/subcommand/mesh-init/command.go#L261) by consul-ecs in consul service metadata during service-registration. This solves our problem. 
Added the reproduction steps for your reference below for the same.


This is most balanced way to meet your use case without much code changes from engineering side and; suppose when you override task definition with newer dataplane image, you have to just update/pass the version fields in`var.consul_service_meta`.

If the consensus is made on this approach, I will add the default values in terraform-module repo for `var.consul_service_meta` and add validation on top of it wrt to `var.consul_dataplane_image`.

Please let me know your thoughts on this and lets close this one. 

--------------------
### Reproduction. 
I had passed dataplane version via `tf-module` `meta` field and also as an environment variable (though i had just logged the env passed variables in `consul-ecs`). 
You can see the final AWS task definition environment in the attached image (meta passed is underlined in blue and env passed underlined in purple). 
Also, same can be reflected in consul servers service as well (refer third image) **that exact same key's which we passed in meta field are present in consul service metadata post consul-ecs service registration**.

```
locals {
  // Define the Consul ECS config file contents.
  serviceExtra = lookup(var.consul_ecs_config, "service", {})
  proxyExtra   = lookup(var.consul_ecs_config, "proxy", {})
  ...
  
  #[versions passed in meta field]
  service_meta = merge(
    var.consul_service_meta,
    {
      "ecs-version"       = var.consul_ecs_image_version
      "dataplane-version" = var.consul_dataplane_image_version
    }
  )

  config = {
    consulLogin = merge(local.consulLogin, local.loginExtra)
    service = merge(
      {
        name      = local.service_name
        tags      = var.consul_service_tags
        port      = var.port
        meta      = local.service_meta      #here
        namespace = var.consul_namespace
        partition = var.consul_partition
      },
      local.serviceExtra
    )
    ...
    healthSyncContainers = local.defaulted_check_containers
    ...
    transparentProxy       = merge(local.tProxyExtra, local.transparentProxy)
    
    #[versions passed in env]
    ecsVersion       = var.consul_ecs_image_version
    dataplaneVersion = var.consul_dataplane_image_version
  }

  encoded_config = jsonencode(local.config)
}
```

image

------

Screenshot 2026-09-04 at 6 47 55 AM

v-rosa: Sorry for the delay, I'm now back from PTO.

Nice one, that would work too, we actually already modify the consul_ecs_config_json env_var during deploy time to update the upstreams.

